PR #119, opened 2024-03-29
feat(worker): add new test files and update dependencies

miseyu:
# Overview
- add integration test

## What I've done
- add integration test

## How I tested
Please create the necessary files as follows
Test data is optional

```
test_id: `entity/filter`

### test data
- tests/fixture/testdata/${test_id}
  -  fixture_files

### workflow file
- tests/fixture/workflow/${test_id}.yaml

### testcase
- tests/testcase/${test_id}.rs
```

Test data can be obtained from ram as follows.
```
ram:///fixture/testdata/entity/filter/filter.csv
```


worker/tests/testcase/entity/filter.rs
```diff
@@ -0,0 +1,38 @@
+use reearth_flow_action::{ActionValue, Port};
+
+use crate::helper::init_test_runner;
+
+#[tokio::test]
+async fn test_run() {
+    let executor = init_test_runner("entity/filter", vec!["filter.csv"]).await;
```
miseyu, line 7: >> [!NOTE]  
Make sure that init_test_runner is set to test_id and the list of files in test data as arguments.

PR #141, opened 2024-04-08
feat(worker): JSONExtractor

jashanbhullar:
# Overview
JSONExtractor - extract values from json and map to attributes 

## What I've done
- Core functionality
- Tests

## What I haven't done
- Integration tests 
- Reading JSON from URL or string

## How I tested
- Unit tests

## Screenshot
NA 

NA

## Memo
NA

### DIFF
--- a/worker/Cargo.lock
+++ b/worker/Cargo.lock
@@ -1287,6 +1287,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "jsonpath-rust"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


miseyu:
@jashanbhullar 
This is not the functionality we are expecting.
Since an attribute is a column, it is basically an operation on hashmap array data.

jashanbhullar:
> @jashanbhullar This is not the functionality we are expecting. Since an attribute is a column, it is basically an operation on hashmap array data.

So you are saying I need to check whether the ActionValue(input) is an ActionValue::Map or not?

miseyu:
The runtime has changed, close it.

PR #157, opened 2024-04-18
feat(worker): format XML

kirisaki:
# Overview

XMLFormatter

## What I've done

- Core functionally
- Tests

## What I haven't done

- Integrated tests

## How I tested

- Unit test

## Screenshot

NA

NA

## Memo

NA

### DIFF
--- a/worker/crates/action-universal/src/xml/formatter.rs
+++ b/worker/crates/action-universal/src/xml/formatter.rs
@@ -1,3 +1,4 @@
+use reearth_flow_action_log::action_log;
 use reearth_flow_xml::{
     parser::read_xml,
     traits::{Element, Node},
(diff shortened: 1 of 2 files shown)


worker/crates/action-universal/src/xml/formatter.rs
```diff
@@ -0,0 +1,169 @@
...
+                                *src = match {
+                                    || {
+                                        read_xml(src)?
+                                            .first_child()
+                                            .ok_or(reearth_flow_xml::error::Error::WrongDocument)?
+                                            .to_xml()
+                                    }
+                                }() {
+                                    Ok(x) => x,
+                                    Err(_) => src.to_string(),
```
miseyu, line 42: In case of this error, please write in the log
reference: 
https://github.com/reearth/reearth-flow/blob/main/worker/crates/action-universal/src/xml/fragmenter.rs#L193-L197

miseyu:
@kirisaki 
Commented! Please check it.

worker/crates/action-universal/src/xml/formatter.rs
kirisaki: Do we need to let it fail whole when there is an error? Or do we just have to log?

miseyu: @kirisaki 
> Do we need to let it fail whole when there is an error? Or do we just have to log?

Just log it once and you are good to go!
Actually, it needs to be routed to a rejected port, but we'll deal with that later.

kirisaki: How do I set parameters `span` and `logger` in `action_log!()`?

miseyu: Pass what is in the ActionContext as is.

kirisaki: I pushed changes. Please check it.

miseyu (approved):
LGTM

PR #161, opened 2024-04-22
feat(ui): Table Component

jashanbhullar:
- Checkbox component
- - Input component
- - Close panel only when clicked on the same selected panel
- - Basic table
- - Datatable example
# Overview
Table component for the UI.

## What I've done
- Basic Checkbox 
- Basic Input 
- Basic table 
- Datatable (with all the controls) example

## What I haven't done 
- Maybe more refactor  

## How I tested
- Storybook and sample data

## Screenshot
NA 

- Implementation scope 
  - Covered the basics 
- Things possibly still missing are:
  - row numbers 
  - search in specific column 
- @tanstack/react-table is very extensive - can add a huge number of options 

## Memo
NA


ui/src/providers/i18n/locales/ja.json
```diff
@@ -46,6 +46,15 @@
   "Settings section 1": "",
   "Workspaces settings": "",
   "Workflow settings": "",
+  "Home": "",
```
KaWaite, line 49: Any idea why ja.json is the only locales file accompanying en.json? 
We have french, spanish and chinese files as well. With `yarn i18n` it should update all of them 🤔

ui/src/features/BottomPanel/components/DataTable/index.tsx
```diff
@@ -1,40 +1,91 @@
...
+  },
+  {
+    accessorKey: "BRIGHT_TI4",
+    header: "BRIGHT_TI4",
+  },
+  {
+    accessorKey: "BRIGHT_TI5",
+    header: "BRIGHT_TI5",
+  },
+  {
```
KaWaite, line 46: This is all boiler so can you at least add a TODO comment stating as much? 
What this DataTable component should end up doing is either receiving processed columns + data (columns based on the data), or receiving data, and based on the data generating the columns.
If you can write that quickly please do, but just a comment is fine now since the main thing is the UI at the moment 🙏🏻

ui/src/features/BottomPanel/components/Console/index.tsx
```diff
@@ -1,6 +1,6 @@
 const LogConsole: React.FC = () => {
   return (
-    <div className="bg-zinc-900 text-yellow-600 text-xs h-[210px] w-[100%] overflow-scroll rounded-md p-1">
+    <div className="bg-zinc-900 text-yellow-600 text-xs h-[400px] w-[100%] overflow-scroll rounded-md p-1">
```
KaWaite, line 3: I think h-full is better here so that it takes up its parent's height (the BottomPanel, where you are defining the height)

ui/src/components/collapsiblePanels/HorizontalPanel.tsx
```diff
@@ -42,20 +42,29 @@ const HorizontalPanel: React.FC<HorizontalPanelProps> = ({
...
+      // Switch panels
       setSelected(content);
+
+      // Open if closed
       if (!isOpen) {
         handleToggle();
       }
     },
-    [isOpen, handleToggle],
+    [isOpen, handleToggle, selected],
```
KaWaite, line 67: Something like this is maybe clearer (this way I think the if statements are done in a way that the code speaks for itself and comments aren't needed)
```
if (content.id !== selected?.id) {
  setSelected(content);
  if (!isOpen) {
    onToggle?.(true)
  }
} else {
  onToggle?.(!isOpen)
}
```
and then remove the handleToggle since its not really necessary in my mind.

ui/src/components/buttons/ButtonWithTooltip/index.tsx
```diff
@@ -17,9 +17,7 @@ const ButtonWithTooltip: React.FC<ButtonWithTooltipProps> = ({
 }) => (
   <Tooltip>
     <TooltipTrigger asChild>
-      <div className="h-full">
-        <Button {...props}>{children}</Button>
-      </div>
+      <Button {...props}>{children}</Button>
```
KaWaite, line 20: Please be sure this change doesn't break UI where ButtonWithTooltip is being used already 🙏🏻

ui/src/components/Table/index.tsx
```diff
@@ -0,0 +1,2 @@
+export * from "./TableComponents";
```
KaWaite, line 1: This is a bit different from other /components. Each index.tsx file of a component in @flow/components is a single component, but here you have 2 components under Table/ being exported.
Please move TableComponents to @flow/components/Table as the index, and then create a DataTable either in components/ (if its generic enough) or in to @flow/features/BottomPanel/components/DataTable directly

ui/src/components/Table/DataTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { useT } from "@flow/providers";
+
+import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from ".";
+
+interface DataTableProps<TData, TValue> {
+  columns: ColumnDef<TData, TValue>[];
+  data: TData[];
+  selectColumns?: boolean;
+  showFiltering?: boolean;
+}
```
KaWaite, line 30: Is there a Prop directly useable in react-table?

ui/src/components/Table/DataTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+  columns,
+  data,
+  selectColumns = false,
+  showFiltering = false,
+}: DataTableProps<TData, TValue>) {
+  const t = useT();
+  const [sorting, setSorting] = useState<SortingState>([]);
+  const [columnVisibility, setColumnVisibility] = useState<VisibilityState>({});
+  const [rowSelection, setRowSelection] = useState({});
+  const [globalFilter, setGlobalFilter] = useState("");
```
KaWaite, line 42: If any of these are being managed already by react-table, and accessible through useReactTable, I don't think we need to keep state here as well

ui/src/providers/i18n/locales/ja.json
jashanbhullar: I didn't notice. I have updated the code.

ui/src/features/BottomPanel/components/DataTable/index.tsx
jashanbhullar: Added the comment for clarity.

ui/src/features/BottomPanel/components/Console/index.tsx
jashanbhullar: I was also looking into this. I want to work on this in the Maximize view PR, is that okay?

ui/src/components/collapsiblePanels/HorizontalPanel.tsx
jashanbhullar: Updated the function code. Thanks

ui/src/components/buttons/ButtonWithTooltip/index.tsx
jashanbhullar: It's used at 3 places. I have verified that the UI hasn't changed after this change.

ui/src/components/Table/DataTable.tsx
jashanbhullar: I followed the guide from here: https://ui.shadcn.com/docs/components/data-table

As I understand this, react-table does expose all the functionality but the state still has to be managed at the component. This can be quite generic in our case. 
For example - columnVisibilty we can just the `selectColumns` to true and we can use the button to select which columns to show. The alternative is we pass the `[columnVisibility, setColumnVisibility]` state into the table. 

Opinion:
Keeping it simple for now should be okay, can extend if needed once we get better understanding on how the user will use this and what kind of data will be there.

ui/src/components/Table/DataTable.tsx
jashanbhullar: Explained here: https://github.com/reearth/reearth-flow/pull/161#discussion_r1602678773

ui/src/components/Table/index.tsx
jashanbhullar: Updated to create the DataTable as a separate component

KaWaite (approved):

PR #207, opened 2024-05-22
feat(ui): Auth0 Integration

jashanbhullar:
# Overview
Integrating auth0

## What I've done
- Integrated based on reearth

## What I haven't done
- Refactor
- Resolve comments (what is multi tenant? Is it needed in the flow codebase?) 
- Logout functionality? 
- User avatar integration

## How I tested
- Visually

- Tanstack router is different from react-router. Do we need to add authentication per page? 

## Memo
- Tanstack Router is different from react-router. Need to understand it for proper implementation


ui/src/config/index.ts
```diff
@@ -40,3 +44,6 @@ export default async function loadConfig() {
 export function config(): Config {
   return window.FLOW_CONFIG ?? {};
 }
+
+// Export everything from authInfo
```
KaWaite, line 48: Don't need this since the code speaks for itself

ui/src/features/Dashboard/components/Nav/components/UserNavigation.tsx
```diff
@@ -17,17 +18,19 @@ import { useDialogType } from "@flow/stores";
...
+  const { logout: handleLogout, user } = useAuth();
+
   return (
     <DropdownMenu>
       <DropdownMenuTrigger>
         <div className="flex gap-2 mr-2">
           <Avatar className="h-8 w-8">
-            <AvatarImage src="https://github.com/KaWaite.png" />
-            <AvatarFallback>KW</AvatarFallback>
+            <AvatarImage src={user?.picture ? user.picture : "https://github.com/KaWaite.png"} />
```
KaWaite, line 28: As a fallback can you replace my thumbnail perhaps with a U for user (to keep inline with the avatar fallback and username)

ui/src/lib/auth/authHook.ts
KaWaite, line 1: Can you rename this to types.ts

KaWaite (approved):

ui/src/lib/auth/authHook.ts
jashanbhullar: I was thinking the same thing :laughing:

jashanbhullar: I actually wanted to check if we can use the `.d.ts` files for types :thinking: 
I am not sure how the difference when tsCompiler looks at it.

PR #222, opened 2024-05-30
feat(ui): Workspace API Integration 

jashanbhullar:
# Overview
Workspace API integration 

## What I've done
- Workspace API integration
- Reads workspace 
- Navigation 
- Create workspace 

## What I haven't done
- Workspace Update & Delete Mutation UI

## How I tested
- Visually

## Screenshot
NA

- Graphql queries


ui/src/types/workspace.ts
```diff
@@ -1,11 +1,9 @@
...
 import type { Project } from "./project";
 
-export type Workspace = {
-  id: string;
-  name: string;
-  description?: string;
-  members: Member[] | undefined;
-  projects: Project[] | undefined;
-  personal?: boolean;
+export type Workspace = Pick<WorkspaceGraphqlType, "id" | "name" | "personal"> & {
```
KaWaite, line 6: Like we were discussing, we want to abstract gql away from code beyond lib/gql. So this should be a lib agnostic type that we move to inside the api code

ui/src/lib/gql/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./queries";
 export * from "./provider/";
+export * from "./__gen__/graphql";
```
KaWaite, line 3: We decided that this wasn't a good idea because of type conflicts, no? 🤔

ui/src/lib/gql/queries/user.ts
```diff
@@ -16,13 +15,14 @@ graphql(`
   }
 `);
 
+export enum UserQueryKeys {
+  GetMe = "getMe",
+}
```
KaWaite, line 20: Is this not available from the generated code?

ui/src/lib/gql/queries/user.ts
```diff
@@ -16,13 +15,14 @@ graphql(`
...
+
 export const useMeQuery = () => {
   const graphQLContext = useGraphQLContext();
-  const { data, ...rest } = useQuery({
-    // TODO: Use static keys rather than strings. Export the keys as well
-    queryKey: ["getMe"],
-    queryFn: async () => graphQLContext?.GetMe(),
+  return useQuery({
+    queryKey: [UserQueryKeys.GetMe],
+    queryFn: () => graphQLContext?.GetMe(),
```
KaWaite, line 26: I thought the `queryFn` needs to be async 🤔

ui/src/lib/gql/queries/workspace.ts
KaWaite, line 1: Hmmm, seeing it all combined in one file makes me feel like perhaps we should have it all in the same dir, but split in to separate files.
Something like
`src/lib/gql/queries/workspace/hooks`
`src/lib/gql/queries/workspace/types`
`src/lib/gql/queries/workspace/raw` <- not sure what to call where we define the graphql(`....`). Maybe workspace/definitions? hmm...

Also, if we do something like the above, perhaps we can drop the queries dir and just do
`src/lib/gql/workspace/queries`
`src/lib/gql/workspace/mutations`
etc

Any thoughts on any of the above in regards to ease of readability/maintenance/etc?

ui/src/lib/gql/queries/workspace.ts
```diff
@@ -0,0 +1,114 @@
...
+  mutation CreateWorkspace($input: CreateWorkspaceInput!) {
+    createWorkspace(input: $input) {
+      workspace {
+        id
+      }
+    }
+  }
+`);
+
+// TODO: Should this a fragment in GET_ME?
```
KaWaite, line 17: I think having a Workspace fragment would be helpful here. Then you can use it in all of these queries and mutations so you can consistently get the same object back and update them all at once if the backend api changes

ui/src/features/Dialog/components/workspace/useWorkspace.tsx
```diff
@@ -0,0 +1,19 @@
+import { StarIcon } from "@radix-ui/react-icons";
```
KaWaite, line 1: Please try to use phosphorus icons, not radix. Radix is very limited compared to phsphoros.

ui/src/features/Dialog/components/workspace/AddWorkspace.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { ContentSection } from "../ContentSection";
+
+const AddWorkspace: React.FC = () => {
+  const t = useT();
+  const [name, setName] = useState<string | undefined>();
+  const [buttonDisabled, setButtonDisabled] = useState(false);
+  const [, setDialogType] = useDialogType();
+  const navigate = useNavigate();
+  const [showError, setShowError] = useState(false);
+  const [createdWorkspaceId, setCreatedWorkspaceId] = useState<string | undefined>(undefined);
```
KaWaite, line 23: Utilizing this adds a lot of code that isn't really necessary. Instead of setting this state, just to handle a navigation in a useEffect, I think its simpler and better to just await the mutation inside the handleClick function, and then redirect if there is a workspace.id.

ui/src/lib/gql/queries/workspace.ts
```diff
@@ -0,0 +1,114 @@
...
+export enum WorkspaceQueryKeys {
+  GetWorkspace = "getWorkspace",
+}
+
+type mutationInput = {
+  onSuccess?: () => void;
+  onError?: () => void;
+};
+
+export const useCreateWorkspaceMutation = ({
```
KaWaite, line 58: Hmm something about all the mutations seem not good to use, and/or limited in their usefulness.
I think passing onSuccess and onError isn't needed. We want these custom hooks to reduce duplicate code, so if we can take in the mutation's input, handle the mutation, and send the results (data, status, etc) like we are doing with the useGetWorkspaceQuery, I think the api will be more consistent and easier to use, etc.
This is the way I wrote it in reearth-viz too, so I am a little biased. If you see your approach to mutations better or see a better solution please let me know and we can discuss.

ui/src/features/Dialog/components/workspace/AddWorkspace.tsx
jashanbhullar: You can wait for the mutation but the query automatically refreshes after that and I need to wait for it to finish before navigating. Is there a better way in tanstack?

ui/src/features/Dialog/components/workspace/useWorkspace.tsx
jashanbhullar: In this case `Icon` isn't even used. Can we update the code it make it optional?

ui/src/lib/gql/index.ts
jashanbhullar: I need the types in the `/types`
I can use the directly `__gen__` folder :thinking: 
I am still not sure, I don't think we should export....
to export or not to export... :thinking:

ui/src/lib/gql/queries/user.ts
jashanbhullar: No, not that I can could find. The other plugin had it but it doesn't work with v5. 
ref: 
> [typescript-react-query](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-react-query) would be better but it's not working with v5 at the moment . Link to the [issue](https://github.com/dotansimha/graphql-code-generator-community/issues/586)

ui/src/lib/gql/queries/user.ts
jashanbhullar: Not necessarily, the Tanstack automatically figures it out.

ui/src/lib/gql/queries/workspace.ts
jashanbhullar: Hmm...this makes sense...Especially with the types :thinking: 

I'll make this change

ui/src/lib/gql/queries/workspace.ts
jashanbhullar: Cool. This makes sense to me :+1:

ui/src/lib/gql/queries/workspace.ts
jashanbhullar: Hmm...we need a way to know if the mutation was successful or not, I'll look into something which is less code

ui/src/types/workspace.ts
jashanbhullar: Hmm..yess. As you said above, I'll move the types in the gql folder

ui/src/features/Dialog/components/workspace/AddWorkspace.tsx
jashanbhullar: I'll see if I can do better

ui/src/lib/gql/index.ts
KaWaite: Ah maybe I left this up to you... I guess personally I feel like knowing something is generated has its merits, so making imports need to specify the __gen__ file is good. If we don't do that, making the __gen__ file for generated code is a little pointless and we should generate directly in the lib/gql dir.
Make sense? Agree disagree?

ui/src/lib/gql/queries/user.ts
KaWaite: Ah yeah this was one of the "we will improve it when we can" things. Gotcha

ui/src/lib/gql/queries/user.ts
KaWaite: I could be wrong, but I think if we wrap an async func in a normal func where we use that func will not await the child. Which we want.
Also, [the docs](https://tanstack.com/query/v4/docs/framework/react/guides/query-functions) show making the queryFn async so should stick with the recommended way

ui/src/features/Dialog/components/workspace/useWorkspace.tsx
KaWaite: Ah in that case maybe just leave this as is. I think the Dialog feature will be simplified so it doesn't have the tabular functionality, so these icons will be deleted too.
Once you read this you can resolve this convo 👍🏻

ui/src/lib/gql/queries/workspace.ts
KaWaite: At least with apollo, the mutation would return `data` and then other info like whether it was successful or not, etc, which is why we return in the custom hook usually `{ data, ...rest }`

ui/src/features/Dialog/components/workspace/AddWorkspace.tsx
KaWaite: We have global state for currentWorkspace, so for starters could just await, and then when we get results we assign it to currentWorkspace state and, navigate.

ui/src/types/workspace.ts
jashanbhullar: With the new dir structure, this is fixed

ui/src/lib/gql/index.ts
jashanbhullar: Makes sense, not exporting it. It should stay abstracted away in the `gql` directory

ui/src/lib/gql/queries/user.ts
jashanbhullar: The logic is you can return a Promise and it'll await that. If it's just the fn we don't need to write `async`, in the doc second example it's not using `async`

ui/src/lib/gql/queries/workspace.ts
jashanbhullar: This is done

ui/src/features/Dialog/components/workspace/useWorkspace.tsx
jashanbhullar: I have updated the icon still. Resolving this!

ui/src/features/Dialog/components/workspace/AddWorkspace.tsx
jashanbhullar: I figured it out. Need to await the refetch. 
Also optimistic cache or UI update won't work in this scenario because:
- We need the ID of the workspace 
- ID will be used to navigate 

We can only get the ID from the backend. I'll keep the optimistic UI and cache in mind for future.

ui/src/lib/gql/queries/workspace.ts
jashanbhullar: This is done. Using `useEffect` now 
```
  useEffect(() => {
    if (createWorkspace.isIdle || createWorkspace.isPending) return;

    if (createWorkspace.isError) {
      setShowError(true);
      setButtonDisabled(false);
    } else if (createWorkspace.isSuccess) {
      const workspaceId = createWorkspace.data?.createWorkspace?.workspace.id;
      setButtonDisabled(false);
      setShowError(false);
      setDialogType(undefined);
      navigate({ to: `/workspace/${workspaceId}` });
    }
  }, [createWorkspace, setShowError, setButtonDisabled, setDialogType, navigate]);

  const handleClick = () => {
    if (!name) return;
    createWorkspace.mutate({ input: { name } });
    setButtonDisabled(true);
  };
```

ui/src/features/Dashboard/components/Nav/components/WorkspaceNavigation.tsx
```diff
@@ -9,22 +9,23 @@ import {
...
-import { workspaces } from "@flow/mock_data/workspaceData";
-import { useCurrentProject, useCurrentWorkspace } from "@flow/stores";
+import { useCurrentWorkspace } from "@flow/stores";
 import { Workspace } from "@flow/types";
 
 const WorkspaceNavigation: React.FC = () => {
-  const [currentWorkspace, setCurrentWorkspace] = useCurrentWorkspace();
-  const [, setCurrentProject] = useCurrentProject();
-  const navigate = useNavigate({ from: "/workspace/$workspaceId" });
+  const [currentWorkspace, _] = useCurrentWorkspace();
```
KaWaite, line 18: You can just write `const [currentWorkspace] ....`. The others not being used is inferred.

ui/src/features/Dashboard/components/Nav/components/WorkspaceNavigation.tsx
```diff
@@ -9,22 +9,23 @@ import {
...
 
   const handleWorkspaceChange = (workspace: Workspace) => {
-    setCurrentProject(undefined);
-    setCurrentWorkspace(workspace);
     navigate({ to: `/workspace/${workspace.id}` });
   };
 
+  const { data } = useGetWorkspaceQuery();
+
+  const workspaces = data?.me?.workspaces;
```
KaWaite, line 27: I think querying and setting workspaces (or others) should hopefully only be in one place per query per page.
This might not be possible later on, but I feel like for objects that won't change often, like workspace or user, we should fetch this data as high up as possible and assign it to the global state. Then, use the global state in child components like WorkspaceNavigation. That way we have state initialization (query + global state set) in one place, which simplifies and synchronizes everything.
Make sense?

ui/src/types/workspace.ts
KaWaite: Sorry I think you missed a key point. You currently are using a generated gql type as the base for the Workspace type (and you have deleted the Workspace app type inside of src/types/.
This isn't ideal since that means we will be indirectly relying on gql inside the app. 
Instead, we should keep the types/workspace.ts, and inside of the lib/gql api hooks convert the data (which would be of the generated gql Workspace type) to an object that meets the app type Workspace.

jashanbhullar: So you are saying concert from GQL Workspace Type to Workspace type in UI ? 
That makes sense.

ui/src/features/Dashboard/components/Nav/components/WorkspaceNavigation.tsx
```diff
@@ -9,20 +9,20 @@ import {
...
   const [currentWorkspace, setCurrentWorkspace] = useCurrentWorkspace();
-  const [, setCurrentProject] = useCurrentProject();
-  const navigate = useNavigate({ from: "/workspace/$workspaceId" });
+  const { workspaces } = useGetWorkspace();
+  const navigate = useNavigate();
 
-  const handleWorkspaceChange = (workspace: Workspace) => {
-    setCurrentProject(undefined);
+  const handleWorkspaceChange = async (workspace: Workspace) => {
+    const route = window.location.pathname;
```
KaWaite, line 23: react router has a useParams hook that you can use to get route information.

jashanbhullar: I need the full route and I have the workspace ID from the currentWorkspace state. 
This is not correct in 100% of the places, it'll fail in some cases like when the 
/workspacesId/project/projectId is open 
For a proper implementation  I want to add a check in the routes/[route].tsx as well.

ui/src/lib/gql/workspace/hook.ts
```diff
@@ -0,0 +1,106 @@
...
+type GetWorkspace = {
+  workspaces: Workspace[] | undefined;
+  isLoading: boolean;
+} & CommonReturnType;
+
+type DeleteWorkspace = {
+  workspaceId: string | undefined;
+} & CommonReturnType;
+
+export const useWorkspaceApi = () => {
```
KaWaite, line 32: `useWorkspace`. (later, useUser, useProject, useXxxxx)

ui/src/lib/gql/workspace/hook.ts
```diff
@@ -0,0 +1,106 @@
...
+} & CommonReturnType;
+
+type GetWorkspace = {
+  workspaces: Workspace[] | undefined;
+  isLoading: boolean;
+} & CommonReturnType;
+
+type DeleteWorkspace = {
+  workspaceId: string | undefined;
+} & CommonReturnType;
```
KaWaite, line 30: Move in to types.ts

KaWaite (approved):

PR #227, opened 2024-06-03
feat(ui): Project API integration

jashanbhullar:
# Overview
Project API integration

## What I've done
- graphql queries 
- hook for executing queries 
- Get/Create/Delete project hooks 

## What I haven't done
- Update Project - Check `Notes` section 

## How I tested
- Manually

- graphql queries
- hook implementation 
- react-query implementation

### Notes:
- Dialog component should be able to return a `boolean` value for `confirm` dialogs (Need this for `delete` projects/workspaces). Should revisit Dialog implementation before adding more components on it 
- The dialog component doesn't stop the navigation events, user can open a dialog and still change pages with the keyboard (should stop all events not related to the dialog)  

### Improvements:
- No pagination implementation as of now 
- Not sure of the proper return structure of the data (data, meta, ...rest [isLoading, isError...so on]), is this okay? Is the react-query abstracted enough?


ui/src/features/Dashboard/components/MainSection.tsx
```diff
@@ -15,16 +16,21 @@ import {
...
 import type { Project } from "@flow/types";
+import { formatDate } from "@flow/utils";
 
 const MainSection: React.FC = () => {
   const t = useT();
   const [currentWorkspace] = useCurrentWorkspace();
   const [currentProject, setCurrentProject] = useCurrentProject();
   const navigate = useNavigate({ from: "/workspace/$workspaceId" });
+  const { useGetProjects, deleteProject } = useProject();
+  const { projects } = useGetProjects(currentWorkspace?.id as string);
```
KaWaite, line 32: As discussed on Slack, do not typecast a type on a possible undefined.

ui/src/features/Dashboard/components/MainSection.tsx
```diff
@@ -33,7 +39,17 @@ const MainSection: React.FC = () => {
...
+  useEffect(() => {
+    if (!projects) return;
+    projects.forEach(p => {
+      p.workflow = generateWorkflows(1)[0];
+    });
+  }, [projects]);
+
+  const handleDeleteProject = async (id: string) => {
+    await deleteProject(id);
+  };
```
KaWaite, line 52: Keep in mind we will eventually want a warning Dialog for actions like this. (once you read this you can resolve this comment yourself 👍🏻 )

ui/src/lib/gql/project/hook.ts
```diff
@@ -0,0 +1,88 @@
...
+    useQuery({
+      queryKey: [ProjectQueryKeys.GetProjects, workspaceId],
+      queryFn: async () => {
+        // TODO: Should have pagination
+        const data = await graphQLContext?.GetProjects({ workspaceId, first: 20 });
+        if (!data) return {};
+        const {
+          projects: { nodes, ...rest },
+        } = data;
+        return { projects: nodes as Project[], meta: rest };
```
KaWaite, line 38: Don't typecast here either. projects will still have all of the variables that come from the query, but will just be invisible to everyone, which isn't ideal.
Instead, map through nodes and create new objects that match `Project`.
That way everything is consistent and we can be more deliberate what we are sending to the app 🙏🏻

ui/src/features/Dashboard/components/MainSection.tsx
jashanbhullar: Added a todo for reference. Same as workspace

ui/src/features/Dashboard/components/MainSection.tsx
jashanbhullar: I passed it as a prop since it needs to exist before we can render this component.

jashanbhullar: We are setting the state in the parent. I am not sure how you feel about using global state at too many places?

ui/src/lib/gql/project/hook.ts
jashanbhullar: This is Done!

ui/src/features/Dashboard/components/MainSection.tsx
KaWaite: SETTING global state should be only in very deliberate places so its easy to track. 
USING global state I think can, and should, be used liberally. That way we can avoid prop drilling and components are more mobile if we chose to rearrange the UI, etc

KaWaite (approved):

ui/src/lib/gql/project/hook.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export const useProject = () => {
+  const graphQLContext = useGraphQLContext();
+  const queryClient = useQueryClient();
+
+  const createProjectMutation = useMutation({
+    mutationFn: async (input: CreateProjectInput) => {
+      const data = await graphQLContext?.CreateProject({ input });
+      return { project: data?.createProject?.project };
```
KaWaite, line 19: Sorry realized you are passing the gql project here

ui/src/lib/gql/project/hook.ts
```diff
@@ -0,0 +1,92 @@
...
+      queryKey: [ProjectQueryKeys.GetProjects, workspaceId],
+      queryFn: async () => {
+        // TODO: Should have pagination
+        const data = await graphQLContext?.GetProjects({ workspaceId, first: 20 });
+        if (!data) return {};
+        const {
+          projects: { nodes, ...rest },
+        } = data;
+
+        const projects: Project[] = nodes.flatMap(f => (f ? [f] : []));
```
KaWaite, line 39: Sorry realized you are passing the gql project here too

ui/src/lib/gql/project/hook.ts
```diff
@@ -0,0 +1,92 @@
...
+      queryClient.invalidateQueries({
+        queryKey: [ProjectQueryKeys.GetProjects, workspaceId],
+      }),
+  });
+
+  const createProject = async (input: CreateProjectInput): Promise<CreateProject> => {
+    const { mutateAsync, ...rest } = createProjectMutation;
+    try {
+      const data = await mutateAsync(input);
+      return { project: data.project, ...rest };
```
KaWaite, line 62: And here

ui/src/lib/gql/project/hook.ts
```diff
@@ -0,0 +1,92 @@
...
+    } catch (err) {
+      return { project: undefined, ...rest };
+    }
+  };
+
+  const useGetProjects = (workspaceId: string): GetProjects => {
+    const { data, ...rest } = useGetProjectsQuery(workspaceId);
+    return {
+      projects: data?.projects,
```
KaWaite, line 71: And here

KaWaite (changes requested):
Sorry noticed something

ui/src/lib/gql/project/hook.ts
jashanbhullar: image
I am using the one from types!

KaWaite (approved):

PR #240, opened 2024-06-10
feat(ui): Route Param Validation

jashanbhullar:
- - lint fixes
- - adding tanstack react-query eslint
- - better way to format data
- - accidently deleted

# Overview


ui/src/lib/gql/workspace/hook.ts
KaWaite, line 1: In this file, if you can also update what the `getWorkspaces` function is returning like you are doing with projects? (new object)

ui/src/lib/gql/user/hook.ts
```diff
@@ -12,18 +12,15 @@ export const useUser = () => {
...
+    select: data => data?.me,
     staleTime: Infinity,
   });
 
   const getMe = (): GetMe => {
-    const { data: me, ...rest } = getMeQuery;
+    const { data, ...rest } = getMeQuery;
     return {
-      me,
+      me: data,
```
KaWaite, line 23: Same here, can you create a new object that matches the app type? 🙏🏻

ui/.eslintrc.cjs
```diff
@@ -24,13 +28,33 @@ module.exports = {
...
     ],
+    "@tanstack/query/exhaustive-deps": 0,
   },
   overrides: [
     {
       files: ["*.graphql"],
       parser: "@graphql-eslint/eslint-plugin",
       plugins: ["@graphql-eslint"],
       extends: "plugin:@graphql-eslint/operations-recommended",
+      // Source: https://github.com/dimaMachina/graphql-eslint/blob/master/packages/plugin/src/configs/operations-recommended.ts
```
KaWaite, line 39: Delete

ui/src/features/PageWrapper/WorkspaceId.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    return;
+  }, [workspace, setCurrentWorkspace]);
+
+  if (isLoading) return <Loading />;
+
+  if (!workspace)
+    return <NotFoundPage message={`Workspace with id: "${workspaceId}" not found.`} />;
+
+  return <>{children}</>;
```
KaWaite, line 36:
```suggestion
  return children;
```

ui/src/features/PageWrapper/WorkspaceId.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { workspaceId }: { workspaceId: string } = useParams({
+    strict: false,
+  });
+
+  const { useGetWorkspace } = useWorkspace();
+  const { workspace, isLoading } = useGetWorkspace(workspaceId);
+
+  useEffect(() => {
+    if (!workspace) return;
+    setCurrentWorkspace(workspace);
```
KaWaite, line 26: We also don't want to reset global state if currentWorkspace.id === workspace.id

ui/src/features/PageWrapper/ProjectId.tsx
KaWaite, line 1: Hmmm I guess I didn't really think we would make wrappers for each global state. This seems excessive and also duplicates a lot of code.. hmmm
Any ideas? Perhaps one wrapper with two hooks files? Or....hmm

KaWaite: Also, for this PR perhaps this implementation is fine. But should be refactored I think.

ui/src/lib/gql/project/useApi.ts
```diff
@@ -26,7 +34,7 @@ export const useProject = () => {
     }
   };
 
-  const useGetProjects = (workspaceId: string): GetProjects => {
+  const useGetWorkspaceProjects = (workspaceId: string): GetWorkspaceProjects => {
```
KaWaite, line 37: Wasn't useGetProjects meaningful enough? Projects will only ever be workspace projects

ui/src/features/Dashboard/MainSection.tsx
KaWaite, line 1: Any reason you moved these out of the components/ dir? Up until now the general structure of components is 
```
Flow/
---/components OR /features
------/MyComponent
--------- index.tsx
--------- hooks.ts
--------- components/
------------index.ts
------------ChildComponentOfMyComponent.tsx
```
I know it can seem a little excessive, but keeping things consistent at least allows the dev to know what to expect in any given component's dir.

ui/.eslintrc.cjs
jashanbhullar: Done!

ui/src/features/PageWrapper/ProjectId.tsx
jashanbhullar: So initially I did that but it was too much code. Also, I'll need to **conditionally** call the `getProjectByID` checking whether the `projectId` was in the params or not. The split helps with the separation of concerns. 

Also we can do the same for the `tabs` and `runs` as well but didn't think it was that crucial at this point.

ui/src/lib/gql/project/useApi.ts
jashanbhullar: So I added the `getProject` which can be used to get a single project which got me thinking. 
This is more explicit stating that `projects` belong to `workspace` hence the name.

ui/src/features/Dashboard/MainSection.tsx
jashanbhullar: Hmm..Since I moved the main component to the `/pages` directory I thought this would just export the components. Since this breaks the convention, I'll revert this.

jashanbhullar: Changed it back

ui/src/lib/gql/project/useApi.ts
jashanbhullar: And workspace ID needs to be supplied. There's also a query (not implemented yet) which is `GetProjects` where you can pass in the `IDs` and set `Node` to `Project` and get all the projects with that `IDs` it was much clearer in my head to rename it as `workspaceProjects`

KaWaite (approved):

PR #261, opened 2024-06-18
feat(ui): Workspace CRUD

jashanbhullar:
# Overview
Workspace CRUD 
- Update Workspace (just the name)
- Team members CRUD 

## What I've done
- graphql queries 
- method implementations

## What I haven't done
NA

## How I tested
- manually

## Screenshot
NA

NA

## Memo
NA

## Summary by CodeRabbit

- **New Features**
  - Enhanced workspace management with the ability to update workspace names, add, remove, and update members.
  - Introduced new roles for workspace members including "Maintainer".

- **Improvements**
  - Refined user information retrieval and display across various components for better performance and accuracy.
  - Updated the `WorkspaceIdWrapper` and `NotFoundPage` components for improved data handling.
  
- **Bug Fixes**
  - Fixed dependency issues in workspace-related components to ensure proper updates and state management.

- **Refactor**
  - Reorganized role types to use enums instead of string literals for consistency and type safety.
  - Adjusted hooks and state management for cleaner and more maintainable code.

- **Documentation**
  - Improved type definitions and added new types to better support user and workspace functionalities.

- **Chores**
  - Removed redundant code and cleaned up the syntax in various files including the routes and settings components.


ui/src/lib/gql/user/useQueries.ts
```diff
@@ -26,7 +26,23 @@ export const useQueries = () => {
     staleTime: Infinity,
   });
 
+  // Not using react-query because no observers are needed on this
+  const searchUserQuery = async (email: string) => {
```
KaWaite, line 30: It doesn't have to be email, right? It can be username too I thought. If I am right, make email -> queryString or searchTerm

ui/src/features/WorkspaceSettings/components/GeneralSettings.tsx
```diff
@@ -6,25 +6,44 @@ import { useWorkspace } from "@flow/lib/gql";
...
+    setShowError(undefined);
+    if (!currentWorkspace?.id || !workspaceName) return;
+    const { workspace } = await updateWorkspace(currentWorkspace?.id, workspaceName);
+    setLoading(false);
+    if (!workspace) {
+      setShowError("update");
+      return;
+    }
+    // TODO: Not sure if this is correct to set here
+    setCurrentWorkspace(workspace);
```
KaWaite, line 45: I feel like this shouldnt be necessary if we are using graphql and the cache effectively. 
I feel like updating the workspace should automatically update the cache, and thus update wherever the workspace(s) query is used.
If that isn't happening, perhaps we need to either
- update the useQueries code to update the cache specifically, OR
- set up the cache better through more specific queryKeys

ui/src/features/WorkspaceSettings/components/GeneralSettings.tsx
```diff
@@ -6,25 +6,44 @@ import { useWorkspace } from "@flow/lib/gql";
...
+  const { deleteWorkspace, updateWorkspace } = useWorkspace();
   const navigate = useNavigate();
-  const [showError, setShowError] = useState(false);
+  const [showError, setShowError] = useState<Errors | undefined>(undefined);
+  const [workspaceName, setWorkspaceName] = useState(currentWorkspace?.name);
+  const [loading, setLoading] = useState(false);
 
   const handleDeleteWorkspace = async () => {
-    setShowError(false);
+    setLoading(true);
```
KaWaite, line 21: Sholdn't be get a `loading` or `isLoading` variable from the gql api that we can use instead of locally setting state? 🤔

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
```diff
@@ -3,87 +3,109 @@ import { useState } from "react";
...
+    { id: Role.Writer, title: t("Writer") },
   ];
 
   const members =
-    (currentWorkspace?.members?.filter(
-      m => "userId" in m && (currentFilter !== "all" ? m.role === currentFilter : true),
-    ) as UserMember[]) ?? [];
+    (
+      currentWorkspace?.members?.filter(
+        m => "userId" in m && (currentFilter !== "all" ? m.role === currentFilter : true),
```
KaWaite, line 42:
```suggestion
        m => "userId" in m && (currentFilter === "all" || m.role === currentFilter),
```

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
```diff
@@ -3,87 +3,109 @@ import { useState } from "react";
...
+      ) as UserMember[]
+    ).sort((a, b) => a.user?.name.localeCompare(b.user?.name ?? "") ?? 0) ?? [];
```
KaWaite, line 44: Personally I feel like until (unless?) we implement an option for alphabetized sorting (A-Z and Z-A) this might be confusing when you add a member and they don't get added to the bottom. Especially if the list is longer than 4 or 5 members.

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
```diff
@@ -105,26 +127,36 @@ const MembersSettings: React.FC = () => {
...
+                    ))}
+                  </DropdownMenuContent>
+                </DropdownMenu>
+                <Button
+                  className="flex-1 h-[25px]"
+                  size="sm"
+                  variant="outline"
+                  onClick={() => handleRemoveMembers(m.userId)}>
+                  {t("Remove")}
+                </Button>
```
KaWaite, line 154: I think lots will change in the coming months around roles as we update reearthx for more advanced role setting, but just to keep in your mind (if you have a good temp solution let me know) that we need to be careful who can do what action around this.
Basic musts are:
- non-Owners cannot change roles of others or remove them
- If its your personal workspace, you cannot change anything here
- etc

About the personal workspace, perhaps we just hide this route when its the currentWorkspace.

ui/src/lib/gql/workspace/useQueries.ts
```diff
@@ -14,35 +20,48 @@ export const useQueries = () => {
...
     },
   });
 
   const useGetWorkspacesQuery = () =>
     useQuery({
       queryKey: [WorkspaceQueryKeys.GetWorkspaces],
       queryFn: () => graphQLContext?.GetWorkspaces(),
-      select: data => data?.me?.workspaces.map(w => createNewWorkspaceObject(w)),
+      select: data =>
+        data?.me?.workspaces.map(w => createNewWorkspaceObject(w)).flatMap(w => (w ? [w] : [])),
```
KaWaite, line 64: flatMap should get rid of all null/undefined items so this should be enough
```suggestion
        data?.me?.workspaces.flatMap(w => createNewWorkspaceObject(w)),
```

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
jashanbhullar: Oh yea, sorry I changed it...i was checking something.

ui/src/features/WorkspaceSettings/components/GeneralSettings.tsx
jashanbhullar: the cache updates okay but workspace doesn't reference cache, it copies the workspace from it. So even after updating the cache, it doesn't show up in the `currentWorkspace`

ui/src/lib/gql/workspace/useQueries.ts
jashanbhullar: I know it should work but Typescript seems to disagree. I tried more things but to no avail. with `as` keyword it does work.

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
jashanbhullar: Yes, can do that with personal workspace. We haven't started interated `role based access` on flow, once we have to clear set of actions a user can and can't do, we can build that into the UI.

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
jashanbhullar: Hmm...true. So I was facing an issue where I update the user role and it'll come on top of the list so I added this automatic sorting...but you are right, What do you suggest we should do?

ui/src/features/WorkspaceSettings/components/GeneralSettings.tsx
jashanbhullar: That only works for `queries`. for mutations we are using `mutateAsync` which returns a promise. We can use the `isLoading` if we use the `mutate` function but then it'll go inside a hook and not a normal function.

ui/src/lib/gql/user/useQueries.ts
jashanbhullar: Me and Piyush discussed and we decided it's best to use email since multiple users can exist with the the same username. 
The PR to be merged is open below. 
https://github.com/reearth/reearthx/pull/41

KaWaite: Oh okay. Awesome

ui/src/features/WorkspaceSettings/components/GeneralSettings.tsx
KaWaite: Does this apply even if you use useCallback/useMemo? 🤔

ui/src/features/WorkspaceSettings/components/GeneralSettings.tsx
KaWaite: Ok. Just keep in mind places where you are doing this, since when we add optimistic updates this probably will become unnecessary 🙏🏻 
When you read this you can resolve this convo yourself 👍🏻

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
KaWaite: Hmm I wonder if this is something inconsistent happening on the backend.... was it always on top? Or on refresh the order was correct? 
It might be a caching issue since the members list is a nested object, perhaps tanstack query's caching is messing up somehow.
Let me know if refreshing fixes the order (p.s. how are you testing this locally?)

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
KaWaite: This is just something while testing we must avoid doing or else we will break workspace access

ui/src/lib/gql/workspace/useQueries.ts
KaWaite: Wasn't the introduction of flatMap a couple PRs ago the solution to the typescript issue? 🤔

ui/src/features/WorkspaceSettings/components/GeneralSettings.tsx
jashanbhullar: I haven't used `useCallback/useMemo` in reference to setCurrentWorkspace and currentWorkspace , let me try this.

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
jashanbhullar: Manually adding the user and setting the new workspace as current Workspace

jashanbhullar: i'll update the comment with my findings

ui/src/lib/gql/workspace/useQueries.ts
jashanbhullar: Yes it does but that assumes the `createNewWorkspaceObject` will return only `Workspace` but I updated that it can return `undefined` as well so it simplified the code in other places
flatmap would work only if the return of the function is `workspace`

ui/src/features/WorkspaceSettings/components/GeneralSettings.tsx
jashanbhullar: image
this fixed it, since id wasn't the same it didn't set the currentworkspace again

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
jashanbhullar: I am using the order from the server and it works okay except when changing the roles

KaWaite (approved):

PR #275, opened 2024-06-24
ci: use workflow_call instead of workflow_dispatch

pyshx:
# Why
This PR attempts to fix an issue observed during deployment nightly version of `api` and `ui`.

# How
The problem was that the build workflows were designed to be run as standalone workflows triggered by `workflow_dispatch` events, not as reusable workflows called from another workflow.

To fix this and make these workflows work as reusable workflows, you'll need to make some modifications:
* For both workflows, change the trigger from workflow_dispatch to workflow_call and define the inputs.
* Replace all instances of github.event.inputs with just inputs.
* Modify the if conditions to use the passed inputs instead of relying on github context.

## Summary by CodeRabbit

- **Chores**
  - Standardized input handling and conditionals in build and deploy workflows for both API and UI components. This improves consistency and reliability of the deployment process.

### DIFF
--- a/.github/actions/build_deploy_api/action.yml
+++ b/.github/actions/build_deploy_api/action.yml
@@ -1,4 +1,4 @@
-Copyname: build-api
+name: build-api
 on:
   workflow_call:
     inputs:
--- a/.github/actions/build_deploy_api/action.yml
+++ b/.github/actions/build_deploy_api/action.yml
@@ -1,26 +1,28 @@
-name: build-api
-description: Builds the Backend API and Its Docker Image
-inputs:
-  sha_short:
-    description: 'sha_short'
-    required: true
-  new_tag:
-    description: 'new_tag'
-    required: true
-  new_tag_short:
-    description: 'new_tag_short'
-    required: true
-  name:
-    description: 'name'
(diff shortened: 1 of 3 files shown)


.github/actions/build_deploy_api/action.yml
```diff
@@ -1,26 +1,28 @@
...
+        type: string
+      sha:
+        required: true
+        type: string
+
 jobs:
   build-api:
     name: Build and release api
     runs-on: ubuntu-latest
-    if: github.event.inputs.name || github.event.inputs.new_tag
```
KeisukeYamashita, line 23: I think you can still make the `new_tag` optional and refer to it so that we can replace the `github.event`.

KeisukeYamashita (approved):
I'm not sure wethear `workflow_dispatches` works with 

```yaml
     uses: ./.github/actions/build_deploy_api
```

Ref: [Creating a reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow)

because it specifies the path to the YAML in the examples.
But let's  give it a shot 🚀

PR #296, opened 2024-07-02
fix(ui): Inconsistent app loading

jashanbhullar:
# Overview
App loading inconsistent bug

## What I've done
- Fixed some code 

## Summary by CodeRabbit

- **New Features**
  - Added detailed error messages for user-related actions in English, Spanish, French, Japanese, and Chinese locales, improving feedback for operations like adding, changing roles, and removing members.

- **Bug Fixes**
  - Improved conditional rendering logic in the `Dashboard`, `NotFoundPage`, `ProjectIdWrapper`, and `WorkspaceIdWrapper` components to enhance user experience during loading and error states.
  - Enhanced member filtering logic in `MembersSettings` to handle null values more gracefully.

- **Refactor**
  - Simplified the `Link` component's `to` prop in the `NotFoundPage` component.
  - Updated type definitions and added new types and enums for better type safety and clarity in the GraphQL schema.

- **Documentation**
  - Updated English locale file with specific error messages for better clarity during user actions.


ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
```diff
@@ -39,14 +39,14 @@ const MembersSettings: React.FC = () => {
   ];
 
   const members = currentWorkspace?.members?.filter(
-    m => "userId" in m && (currentFilter === "all" || m.role === currentFilter),
-  ) as UserMember[];
+    (m): m is UserMember => "userId" in m && (currentFilter === "all" || m.role === currentFilter),
```
KaWaite, line 42: I think this change is misleading. m is not only UserMember, but could be IntegrationMember too. We do the filter to get an array of UserMembers.

ui/src/features/PageWrapper/WorkspaceId.tsx
```diff
@@ -30,7 +30,7 @@ const WorkspaceIdWrapper: React.FC<Props> = ({ children }) => {
 
   if (isLoading) return <Loading />;
 
-  if (!workspace)
+  if (!workspace || !currentWorkspace)
```
KaWaite, line 33: I feel like this will show the NotFoundPage more often than before, since after isLoading becomes false, we will have workspace, but won't have currentWorkspace for another rerender.

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
jashanbhullar: I know, it's better than the previous implementation at the very least which was `as UserMember[]` , I did try to `type narrow` on array to get just the `UserMember` and filter out the `IntegrationMember` but this was the best I could come up with. 
 `"userId" in m` states it's a `UserMember` but typescript doesn't get it until i add `is UserMember`. 
 
 Let me know if I can rewrite this in a better way.

ui/src/features/PageWrapper/WorkspaceId.tsx
jashanbhullar: Hmm...you are right about that. What do you think could be the alternative? I need `currentWorkspace` down the children components and then I'll need to check and add the loading state there.

ui/src/features/WorkspaceSettings/components/MembersSettings.tsx
KaWaite: I think the way it was is the best we can do (there might be a cleverer, more advanced TS way to do it, but I do believe this is a common issue when you make something a type that is two different possible types).
So please revert this change

KaWaite (approved):

PR #298, opened 2024-07-03
feat(api): add Workflow to pkg

KaWaite:
# Overview

## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive methods for constructing, managing, and testing `Workflow`, `Node`, `Edge`, and `Graph` objects in the workflow package.

- **Tests**
  - Added extensive unit tests for `Node`, `Edge`, `Graph`, and `Workflow` objects to ensure reliability and correctness.

- **Refactor**
  - Enhanced the `Builder` struct and associated methods for streamlined workflow construction and management.
  - Improved getter and setter methods for workflow components to bolster encapsulation and control.


api/pkg/workflow/edge_test.go
```diff
@@ -0,0 +1,37 @@
...
+	e := &Edge{}
+	e.SetFromPort("fromPort")
+	assert.Equal(t, "fromPort", e.fromPort)
+}
+
+func TestEdge_SetToPort(t *testing.T) {
+	e := &Edge{}
+	e.SetToPort("toPort")
+	assert.Equal(t, "toPort", e.toPort)
+}
```
pyshx, line 37: In general, for such simple setters, it's often sufficient to test them indirectly through other, more complex methods that use them. This approach provides better test coverage of actual behavior while reducing the overhead of maintaining many small, simple tests.

For simple setters like these, You go with Combine tests i.e Instead of individual tests for each setter, you could have a single test that checks all setters.(Just an example we don't we even need this), we might need something like this if functions get too complicated.

```suggestion
package workflow

import (
    "testing"
    "github.com/stretchr/testify/assert"
)

func TestEdgeSetters(t *testing.T) {
    e := &Edge{}
    e.SetID("testID")
    assert.Equal(t, "testID", e.id, "SetID should correctly set the id field")
    fromNodes := []string{"node1", "node2"}
    e.SetFrom(fromNodes)
    assert.Equal(t, fromNodes, e.from, "SetFrom should correctly set the from field")
    toNodes := []string{"node3", "node4"}
    e.SetTo(toNodes)
    assert.Equal(t, toNodes, e.to, "SetTo should correctly set the to field")
    e.SetFromPort("fromPort")
    assert.Equal(t, "fromPort", e.fromPort, "SetFromPort should correctly set the fromPort field")
    e.SetToPort("toPort")
    assert.Equal(t, "toPort", e.toPort, "SetToPort should correctly set the toPort field")
}
```

api/pkg/workflow/node_test.go
```diff
@@ -0,0 +1,37 @@
+package workflow
```
pyshx, line 1: Same comment as above.

pyshx (changes requested):
Can you add changes from https://github.com/reearth/reearth-flow/pull/303 here? since both are making change to the graphl, it doesn't make sense to merge this and then merge that later....

api/gql/workflow.graphql
```diff
@@ -17,12 +11,9 @@ type NodeType {
...
 type Data {
-  name: String
-  inputs: [Input]!
-  outputs: [Output]!
-  status: Status
-  content: String
-  transformerId: ID
+  inputs: [String]!
+  outputs: [String]!
+  actionId: ID
```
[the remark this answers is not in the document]
KaWaite: Yep!

api/pkg/workflow/edge_test.go
KaWaite: Done

api/pkg/workflow/node_test.go
KaWaite: Done

api/pkg/workflow/workflow_test.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func TestWorkflow_SetGraphs(t *testing.T) {
+	w := &Workflow{}
+	wGraphs := []Graph{{}}
+	w.SetGraphs(wGraphs)
+	if len(w.graphs) != 1 {
+		t.Errorf("expected %v, got %v", wGraphs, w.graphs)
+	}
+}
```
pyshx, line 48: Instead of Setter Functions multiple tesst, Can you write one test? and Can you also add test for defining a workflow struct from a dummy values.

PR #392, opened 2024-07-31
Use anyhow for error handling instead of Box<dyn Error>

kasugamirai:
# Overview


pyshx (commented):
It becomes more messy to investigate errors when `anyhow` is introduced, That's why I choose to use `Box<dyn Error>`. Please outline the reasons why we should use `anyhow`

PR #564, opened 2024-10-09
feat(ui): Right Panel Action Parameters

jashanbhullar:
# Overview
Right panel now loads action parameters for each action

## What I've done
- updated types in a few places
- using schemaform for parameters

## How I tested
- Manually

## Screenshot
image

## Release Notes

- **New Features**
  - Enhanced drag-and-drop functionality for canvas components with improved node creation.
  - Streamlined action filtering and search capabilities in the Actions component and Node Picker Dialog.

- **UI Changes**
  - Simplified parameter editing interface using a schema-driven approach in the Param Editor.

- **Localization Updates**
  - Removed outdated keys from localization files across multiple languages, improving clarity and focus on relevant UI elements.

- **Type Safety Improvements**
  - Updated type definitions for actions and nodes to enforce required properties, enhancing data integrity.

### DIFF
--- a/ui/src/features/Editor/components/Canvas/useDnd.ts
+++ b/ui/src/features/Editor/components/Canvas/useDnd.ts
@@ -56,7 +56,7 @@ export default ({
         id: randomID(),
         position,
         data: {
-          name: "sample",
+          name: "",
           status: "idle",
           locked: false,
           onDoubleClick: onNodeLocking,
--- a/ui/src/features/Editor/components/Canvas/useDnd.ts
+++ b/ui/src/features/Editor/components/Canvas/useDnd.ts
@@ -56,6 +56,7 @@ export default ({
         id: randomID(),
         position,
         data: {
+          name: "sample",
           status: "idle",
           locked: false,
           onDoubleClick: onNodeLocking,
--- a/ui/src/features/Editor/components/Canvas/useDnd.ts
+++ b/ui/src/features/Editor/components/Canvas/useDnd.ts
@@ -56,7 +56,6 @@ export default ({
         id: randomID(),
         position,
         data: {
-          name: "",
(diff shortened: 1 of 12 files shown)


ui/src/features/Editor/components/Canvas/useDnd.ts
```diff
@@ -56,6 +56,7 @@ export default ({
         id: randomID(),
         position,
         data: {
+          name: "sample",
```
KaWaite, line 59: This code isn't boilerplate or temp so "sample" isn't appropriate.
Is there a reason you added this specifically?

jashanbhullar: I made this change in the types 
image

Since node can't exist without name otherwise the endpoint `/actions/{name}` won't work 

Due to this change I had added `sample`  but I changed it to `""` to make more sense, the code after this line always add a `name` field so it's always overriden. (check the image below)

image

ui/src/features/Editor/components/RightPanel/components/ParamEditor/index.tsx
```diff
@@ -54,6 +47,11 @@ const ParamEditor: React.FC<Props> = ({
   const { zoomIn, zoomOut, fitView } = useReactFlow();
   const { isFullscreen, handleFullscreenToggle } = useFullscreen();
 
+  const { useGetActionById } = useAction();
+
+  // nodeMeta.name is always defined, this is for the TS error to go away
```
KaWaite, line 52:
```suggestion
  // For action nodes, nodeMeta.name is always defined. Only actions can open 
  // the ParamsEditor, so this is for the TS error to go away
```

KaWaite (approved):

PR #595, opened 2024-10-27
feat(ui): Subworkflow rename

jashanbhullar:
# Overview
Subworkflow rename

## What I've done
- implemented workflow rename

## What I haven't done
NA

## How I tested
Manually

## Screenshot
I tried adding a `gif` but it's bad quality 

[Screencast from 2024-10-27 20-00-30.webm](https://github.com/user-attachments/assets/bf87dac1-5f3f-4a43-bdb2-026a71a4d5f8)
Screencast from 2024-10-27 20-00-30

NA

## Memo
NA

## Release Notes

- **New Features**
	- Added functionality to rename workflows directly within the UI.
	- Introduced new state management for editing workflow names in the WorkflowTabs component.
	- New localization entries for workflow management in English, Spanish, French, Japanese, and Chinese.

- **Bug Fixes**
	- Improved handling and tracking of workflow renaming actions.

- **Documentation**
	- Updated localization files to include new keys for future translations.


ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -135,9 +135,51 @@ export default ({
...
+
+  const handleWorkflowRename = useCallback(
+    (id: string, name: string) =>
+      undoTrackerActionWrapper(() => {
+        if (!name.trim()) {
+          throw new Error("Workflow name cannot be empty");
+        }
+
+        // Update local state
+        setWorkflows((w) => w.map((w) => (w.id === id ? { ...w, name } : w)));
```
KaWaite, line 147: We should be updating, `data.name` instead and then in the custom node doing `data?.name || name`.

ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -135,9 +135,51 @@ export default ({
...
+            if (node.id === workflowId) {
+              const updatedNode = {
+                ...node,
+                data: { ...node.data, name },
+              };
+              mainWorkflowNodes.delete(index);
+              mainWorkflowNodes.insert(index, [updatedNode]);
+            }
+          });
+        }
```
KaWaite, line 174: I'm not sure we really need to do any index checking (anyways, the local state is already updated so putting the check here would only potentially mess up synchronization with local state and the yDoc).
Also, any reason at this point you can't use handleNodesUpdate? Might be able to simplify code here a bit and reduce similar code

ui/src/lib/yjs/useYWorkflow.ts
jashanbhullar: that's for the yWorkflow nodes, you'll see the same change on 164!

ui/src/lib/yjs/useYWorkflow.ts
jashanbhullar: I removed the index checking. I tried using the handleNodesUpdate but it didn't work probably because the undoActionsHandler loops inside.

ui/src/lib/yjs/useYWorkflow.ts
KaWaite: No, but here you are updating the workflow's name on the main object, not the data.name.

ui/src/lib/yjs/useYWorkflow.ts
KaWaite: Just removing the `undoTrackerActionWrapper` that is wrapping `hadnleWorkflowRename` and then using `handleNodesUpdate` should work (in my mind). 
There is a undo tracker provider attached to the y.doc, and then we use `undoTrackerActionWrapper` to watch specific parts of the y.doc we want to listen to (and the rest should be ignored).Since the rest of handleWorkflowRename has nothing to do with Yjs, this should work as expected

jashanbhullar: As discussed this creates a circular dependency issue

ui/src/lib/yjs/useYWorkflow.ts
jashanbhullar: As discussed this is required

ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -139,9 +139,37 @@ export default ({
...
+
+        for (const node of mainWorkflowNodes) {
+          if (node.id === id) {
+            node.data = {
+              ...node.data,
+              name,
+            };
+          }
+        }
+      }),
```
KaWaite, line 165: This is different from what we discussed I think. Here you are manually modifying the nodes directly, whereas elsewhere we rely on yjs methods by removing and adding nodes again.
I am seeing some performance issues on the PR preview, which might be netlify's fault, but also just to keep updating of yjs state not dissimilar, can you do the same here as the CRUD of nodes and edges? 🙏🏻

PR #663, opened 2024-11-28
feat(ui): improve canvas navigation panel

popy01012:
# Overview
I made LeftPanel instantly reflect the nodes on the current canvas, and I created a new lock that works with LeftPanel.
## What I've done

- Hide Transformers folder when no transformers
- Show notes, batches, subworkflows on left panel
- Give unique ID to distinguish the same node on LeftPanel
- Double click on item to lock it
- make unlock button in LeftPanel
- zoom canvas to selected item on lock

## What I haven't done
Lock still allows certain interactions on some nodes
## How I tested
Manually
## Screenshot
image
image

## Release Notes

- **New Features**
  - Introduced node locking functionality in the editor, enhancing user control over node interactions.
  - Added a new `LockerProvider` to manage locking state across components.
  - Enhanced note editing with real-time updates to content.
  - New icons and improved rendering logic in the LeftPanel for various node types.

- **Localization**
  - Added "Subworkflow" entries in English, Spanish, French, Japanese, and Chinese localization files.

- **Bug Fixes**
  - Improved node creation process to ensure unique names for new nodes.

- **Chores**
  - Cleaned up code by removing commented-out sections and reorganizing props in components.

### DIFF
--- a/ui/src/utils/randomIDshort.ts
+++ b/ui/src/utils/randomIDshort.ts
@@ -1,0 +1,12 @@
+export const randomIDshort = (length = 6): string => {
+  let result = "";
+  const characters =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+  const charactersLength = characters.length;
+
+  for (let i = 0; i < length; i++) {
+    result += characters.charAt(Math.floor(Math.random() * charactersLength));
+  }
+
+  return result;
+};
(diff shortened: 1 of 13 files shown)


ui/src/utils/randomIDshort.ts
KaWaite, line 1: This isn't necessary. We already have `randomID`. randomID's length prop is set to 24 as a default value, but if a value is passed in to the function it'll make an id of that length.

PR #671, opened 2024-12-05
feat(ui): add select lang to AccountUpdateDialog

billcookie:
# Overview
Update to the AccountUpdateDialog to allow users to select a different language from the one they use at signup.
## What I've done
I added the ability for a user to change their selected language. Updated the GetMeQuery and UpdateMe mutation.

## How I tested
Manually
- Functionality 
- Improvements

## Memo
- I believe sometimes users might have the initial lang as und. I have tried to prevent this by falling back on i18n

## Release Notes

- **New Features**
	- Users can now select their preferred language in the Account Update dialog.
	- Language preferences are saved alongside user account updates.
	- Language selection is now available in English, Spanish, French, Japanese, and Chinese.

- **Bug Fixes**
	- Improved error handling for language update failures.

- **Documentation**
	- Updated localization files to support the new language selection feature.


ui/src/routes/workspaces.tsx
```diff
@@ -1,11 +1,25 @@
...
+  const { me } = useGetMe();
+  const selectedLanguage =
+    me?.lang && me.lang !== "und" ? me.lang : i18n.language;
+
+  useEffect(() => {
+    if (selectedLanguage) {
+      i18n.changeLanguage(selectedLanguage);
+    }
+  }, [selectedLanguage]);
+
```
KaWaite, line 21: Is it possible to move this to [here](https://github.com/reearth/reearth-flow/blob/c782baca8b7b74b2adbcff6cab28faa6e183080d/ui/src/lib/i18n/Provider.tsx)?

billcookie: No problem! Will move across now and update.

KaWaite (approved):

PR #679, opened 2024-12-06
fix(ui): Disable Redo & Undo if no actions available

billcookie:
# Overview
Currently the buttons are forever clickable and no ui difference when there is, or isn’t, available undo or redo actions.
What we want is a UI indication that there are no actions available and have it so the user can’t click them either.
## What I've done
Track the length of the undoManagement stack to check if it has any history.
## How I tested
Manually
## Release Notes

- **New Features**
  - Introduced `canUndo` and `canRedo` properties to enhance the functionality of the Toolbox and OverlayUI components, allowing for better control of undo and redo actions.
  - Updated the Editor component to reflect the current state of undo and redo capabilities.
  - Enhanced state management in the application by integrating new properties into the hooks and components.

- **Bug Fixes**
  - Improved the interaction logic for the undo and redo buttons to ensure they accurately reflect the application's state.

- **Documentation**
  - Updated type definitions to include new properties, improving clarity for developers using these components.

### DIFF
--- a/ui/src/lib/yjs/useYjsStore.ts
+++ b/ui/src/lib/yjs/useYjsStore.ts
@@ -37,17 +37,18 @@ export default ({
   const { createDeployment, useUpdateDeployment } = useDeployment();
 
   const handleWorkflowUndo = useCallback(() => {
-    if (undoManager?.undoStack && undoManager.undoStack.length > 0) {
+    const stackLength = undoManager?.undoStack?.length ?? 0;
+    if (stackLength > 0) {
       undoManager?.undo();
     }
   }, [undoManager]);
 
   const handleWorkflowRedo = useCallback(() => {
-    if (undoManager?.redoStack && undoManager.redoStack.length > 0) {
+    const stackLength = undoManager?.redoStack?.length ?? 0;
+    if (stackLength > 0) {
       undoManager?.redo();
     }
   }, [undoManager]);
-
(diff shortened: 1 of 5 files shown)


ui/src/lib/yjs/useYjsStore.ts
```diff
@@ -48,6 +48,16 @@ export default ({
...
 
+  const canUndo = useMemo(() => {
+    const stackLength = undoManager?.undoStack?.length ?? 0;
+    return stackLength > 0;
+  }, [undoManager?.undoStack?.length]);
+
+  const canRedo = useMemo(() => {
+    const stackLength = undoManager?.redoStack?.length ?? 0;
+    return stackLength > 0;
+  }, [undoManager?.redoStack?.length]);
```
KaWaite, line 59: Please use these in handleWorkflowUndo/Redo in their respective conditional statements as well 🙏🏻

KaWaite (approved):

ui/src/lib/yjs/useYjsStore.ts
billcookie: You mean something like this? 

`const handleWorkflowUndo = useCallback(() => {
    const stackLength = undoManager?.undoStack?.length ?? 0;
    if (stackLength > 0) {
      undoManager?.undo();
    }
  }, [undoManager]);`

PR #694, opened 2024-12-13
refactor(ui): synchronized and cleaned node creation

billcookie:
# Overview
Essentially useDnd and nodepickerdialog had the same creation methods for nodes. This has been moved into a useCreateNode hook instead for DRY purposes. 

## What I've done
Moved creation logic which was repetitive from useDnd and nodepickerdialog component into a separate hook called useCreateNode. This makes it easier to maintain in the future for new possible node types.

## How I tested
Manually
- Quality and purpose
- Any further improvements that can be made

### DIFF
--- a/ui/src/features/Editor/components/Canvas/useCreateNode.ts
+++ b/ui/src/features/Editor/components/Canvas/useCreateNode.ts
@@ -1,0 +1,103 @@
+import { XYPosition } from "@xyflow/react";
+
+import { config } from "@flow/config";
+import { fetcher } from "@flow/lib/fetch/transformers/useFetch";
+import { nodeTypes, type Action, type Node, type NodeType } from "@flow/types";
+import { randomID } from "@flow/utils";
+
+import { baseBatchNode } from "./components/Nodes/BatchNode";
+import { baseNoteNode } from "./components/Nodes/NoteNode";
+
+type CreateNodeOptions = {
+  position: XYPosition;
+  type: string;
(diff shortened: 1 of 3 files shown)


ui/src/features/Editor/components/Canvas/useCreateNode.ts
```diff
@@ -0,0 +1,98 @@
...
+        data: { ...node.data, ...baseNoteNode },
+      };
+    default:
+      return null;
+  }
+};
+
+const createNodeFromAction = (action: Action, position: XYPosition): Node => ({
+  ...createBaseNode({ position, type: action.type }),
+  // Needs measured, but at time of creation we don't know size yet.
```
KaWaite, line 49: Can you consolidate params. For two you are passing an object, and then for createNodeFromAction you are passing individual params 🙏🏻

ui/src/features/Editor/components/Canvas/useCreateNode.ts
```diff
@@ -0,0 +1,98 @@
...
+  switch (type) {
+    case "batch":
+      return { ...node, ...baseBatchNode };
+    case "note":
+      return {
+        ...node,
+        data: { ...node.data, ...baseNoteNode },
+      };
+    default:
+      return null;
```
KaWaite, line 43: Instead of returning null, do you think maybe its better just to return the node as-is?

KaWaite (approved):

PR #695, opened 2024-12-13
feat(ui): Batch Node Customization

jashanbhullar:
# Overview
Batch node customization

## What I've done
- basic implementation 
- Schema form 
- Parameter Update

## What I haven't done
- background Color and color are being set but they are not showing up in UI 
- Check `src/features/Editor/components/Canvas/components/Nodes/BatchNode/index.tsx` TODO (not sure why) 
- Issue for why the colour was not showing: React does not always support in line style !important it seems (it does when values like height and pixels etc are passed) and because tailwind is applying important to everything makes it hard to pinpoint how to resolve this. Fix can be found here: https://stackoverflow.com/questions/23074748/important-inline-styles-in-react

## How I tested
- manually

## Screenshot
image

- background Color and color are being set but they are not showing up in UI 

## Summary by CodeRabbit

- **New Features**
  - Enhanced batch node customization with options for custom names, background and text colors.
  - Improved parameter editing for more reliable node configuration with fallback logic.
  - Updated navigation labels (“Previous selection” and “Next selection”) available in multiple languages.

- **Style**
  - Refined node appearance featuring cleaner visuals with updated resizing behavior for a smoother interface.


ui/src/features/Editor/components/Canvas/components/Nodes/utils.ts
```diff
@@ -16,3 +16,20 @@ export const getPropsFrom = (status?: Status) => {
     animated: isAnimated,
   };
 };
+
+export const convertHextoRgba = (hex: string, alpha: number) => {
```
KaWaite, line 20: This could be useful elsewhere, so move it to src/utils

KaWaite (approved):

PR #743, opened 2025-01-07
fix(ui): SubWork Flow Multi Select Deletion

billcookie:
# Overview
Originally subwork flows could only be deleted one by one and multi selecting subwork flows for deletion would not work (error: “Exceeds maximum length”)
## What I've done
- Essentially it seemed originally we could not handle multiple deletions and that when we deleted multiple subwork flows the index would "shift" and not properly realign. This should now be fixed. 
## How I tested
Manually
- Refactors and possible Improvements
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved workflow removal process to handle node and workflow deletion more efficiently.
	- Enhanced logic to prevent index shifting issues during workflow removal.
	- Streamlined the removal logic for better clarity and performance.


ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -147,29 +147,38 @@ export default ({
...
-
-            // Loop over workflow at current index and remove any subworkflow nodes
-            (rawWorkflows[index].nodes as Node[]).forEach((node) => {
+        nodeIds.forEach((nid) => {
+          if (nid === DEFAULT_ENTRY_GRAPH_ID) return;
+          workflowIds.push(nid);
+          const workflow = rawWorkflows.find((w) => w.id === nid);
+          const nodes = workflow?.nodes;
+          // Loop over workflow and remove any subworkflow nodes
+          if (nodes && Array.isArray(nodes)) {
```
KaWaite, line 156: This change iterates on one level deep, and then stops. This is why removeNodes was using itself, to be able to reach ALL subworkflows, no matter how deep.
If you can console.log rawWorkflows and test to make sure when deleting subworkflows that contain subworkflows, that all appropriate workflows are removed on subworkflow node deletion 🙏🏻

ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -173,6 +174,7 @@ export default ({
         setOpenWorkflowIds((ids) =>
           ids.filter((id) => !workflowIds.includes(id)),
         );
+        console.log("workflowIds", rawWorkflows);
```
KaWaite, line 177: Delete please

KaWaite (approved):

PR #756, opened 2025-01-10
feat(ui): Added subflow creation on multi select.

billcookie:
# Overview
Sub workflows could not be created via multi selecting. 
## What I've done
- Added logic to allow sub work flows to be created with multi selection with keyboard short cut "s"
## How I tested
manually
- Possible refactors (particularly with filtering etc)
## Release Notes

- **New Features**
  - Added keyboard shortcut to create a new sub-workflow from selected nodes.
  - Introduced a new keyboard shortcut for adding workflows directly from selected nodes.

- **Localization**
  - Added translations for "Create new sub workflow from selected nodes" in English, Spanish, French, Japanese, and Chinese.

- **Improvements**
  - Enhanced workflow addition functionality to support more complex node relationships.
  - Added new methods for managing workflow states.


ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -42,12 +42,53 @@ export default ({
...
   const handleWorkflowAdd = useCallback(
-    (position?: XYPosition) =>
+    (position?: XYPosition, nodes?: Node[], edges?: Edge[]) =>
       undoTrackerActionWrapper(async () => {
         const workflowId = generateUUID();
-        const workflowName = "Sub Workflow-" + yWorkflows.length.toString();
+        const workflowName = `Sub Workflow-${yWorkflows.length}`;
 
-        const inputRouter = await fetcher<Action>(`${api}/actions/InputRouter`);
+        const nodesByParentId = new Map<string, Node[]>();
```
KaWaite, line 50: Isn't a lot of this logic in useBatch (or elsewhere)?

ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -42,12 +42,53 @@ export default ({
   );
 
   const handleWorkflowAdd = useCallback(
-    (position?: XYPosition) =>
+    (position?: XYPosition, nodes?: Node[], edges?: Edge[]) =>
```
KaWaite, line 45: Adding nodes and edges to here is a little confusing since most places `handleWorkflowAdd` is used only wants position.
Consider a separate function like `handleWorkflowAddFromSelection` or something, while keeping in mind keeping duplicated logic to a minimum.

ui/src/lib/yjs/useYWorkflow.ts
billcookie: We have something similar to this in the batchnode component but not necessarily in the useBatch. The one in the batchnode component largely handles resize logic and direct parent-child relationships and the subworkflow one has some additional complexities. I would be a little bit worried that if we tried to make it so abstracted that it could make it harder to debug in the future as at the moment they are self contained in the places they are used.

KaWaite (approved):

PR #766, opened 2025-01-14
fix(ui): Removed reader add if more than one

billcookie:
# Overview
Reader and Writer were accessible on type and category in Subworkflow and Readers could be created after the existence of one in main work flow

## What I've done
- Removed readers and writers from type, category and keyboard short cuts on subworkflows
- Removed readers from being accessible after the creation of one on main workflow
## How I tested
manually
Possible refactors and improvements
## Summary by CodeRabbit

- **New Features**
	- Added dynamic filtering of actions and tools based on workflow context.
	- Introduced `isMainWorkflow` logic to control UI component behavior.
	- Added a new hook, `useHasReader`, to determine the presence of reader nodes.

- **Improvements**
	- Enhanced node picker functionality to restrict behavior based on workflow context.
	- Refined action and tool display logic in editor components.

- **Technical Updates**
	- Updated API filtering mechanisms to support more granular action selection.
	- Introduced `nodes` parameter in various components and hooks for improved filtering.
	- Added `hasReader` prop to multiple components for better context handling.


ui/src/features/Editor/components/OverlayUI/components/Toolbox/index.tsx
```diff
@@ -13,7 +13,7 @@ import { createRoot } from "react-dom/client";
 
 import { IconButton } from "@flow/components";
 import { useT } from "@flow/lib/i18n";
-import { type NodeType } from "@flow/types";
+import { type NodeType, Node } from "@flow/types";
```
KaWaite, line 16:
```suggestion
import type { NodeType, Node } from "@flow/types";
```

ui/src/features/Editor/components/OverlayUI/components/Toolbox/index.tsx
```diff
@@ -78,6 +79,10 @@ const Toolbox: React.FC<Props> = ({
     if (!isMainWorkflow) {
       return tool.id !== "reader" && tool.id !== "writer";
     }
+
+    if (isMainWorkflow && nodes?.some((node) => node.type === "reader")) {
```
KaWaite, line 83: Passing nodes around and checking some EACH time isn't really desirable. Exactly like `isMainworkflow`, create a single callback that takes in the nodes and checks if it already has a Reader, and then pass the boolean instead.

KaWaite (approved):

PR #871, opened 2025-02-14
fix(ui): Improved Canvas Nav V-2

billcookie:
# Overview

## Summary by CodeRabbit

- **New Features**
  - Enhanced node interactions with double-click behavior that adjusts view and improves selection.
  - Expanded iconography for nodes, offering clearer visual cues.
  - Introduced a function for retrieving node icons based on type.
  - Extended localization support with new labels for additional workflow and batch node features.

- **Bug Fixes**
  - Resolved inconsistencies in selection handling and removed unintended resize behavior.

- **Style**
  - Updated the visual feedback for selected items by switching from a red to a green border.


ui/src/lib/i18n/locales/ja.json
```diff
@@ -49,6 +49,8 @@
   "Type": "タイプ",
   "Loading": "読み込み中",
   "Transformers": "トランスフォーマー",
+  "Subworkflows": "",
+  "Batch Nodes": "",
```
KaWaite, line 53: At least Japanese is necessary

ui/src/features/Editor/useNodeLocker.ts
```diff
@@ -16,19 +16,26 @@ export default ({
   const [locallyLockedNode, setLocallyLockedNode] = useState<Node | undefined>(
     undefined,
   );
+  const isLockingRef = useRef(false);
```
KaWaite, line 19: Why is this ref necessary?

ui/src/features/Editor/hooks.ts
```diff
@@ -158,6 +158,12 @@ export default ({
     [openPanel],
   );
 
+  const handlePanelClose = useCallback(() => {
```
KaWaite, line 161: This shouldn't be necessary. This works in main branch 🤔

ui/src/features/Editor/components/ParamsPanel/index.tsx
```diff
@@ -9,20 +9,19 @@ import { ParamEditor } from "./components";
 type Props = {
   selected?: Node;
   onParamsSubmit: (nodeId: string, data: any) => void;
+  onClose?: () => void;
 };
 
-const ParamsPanel: React.FC<Props> = ({ selected, onParamsSubmit }) => {
-  // This is a little hacky, but it works. We need to dispatch a click event to the react-flow__pane
-  // to unlock the node when user wants to close the right panel. - @KaWaite
+const ParamsPanel: React.FC<Props> = ({
```
KaWaite, line 15: Removing the hacky stuff de-synchronizes the canvas and the app state. By using this hacky code, you are relying on react-flow to make the updates to the node to deselect.

ui/src/features/Editor/useNodeLocker.ts
billcookie: Ahh yes I added it because when trying to use the similar function that sets the selectedId/locallyLocked node. It is set for a split second and then reverted back to undefined, but going through it now and the comment about the close I get what you mean about not using useRef and the onClose. 

I will have to look into it more, but I am not sure why it is being set to undefined so quickly

ui/src/features/Editor/components/LeftPanel/index.tsx
```diff
@@ -28,46 +40,112 @@ const LeftPanel: React.FC<Props> = ({
...
+    }
+  }, [selected, isOpen, getZoom, nodeId]);
 
   useEffect(() => {
     if (!isOpen && selectedTab) {
       setSelectedTab(undefined);
     }
   }, [isOpen, selectedTab]);
 
+  const createTreeDataItem = (type: string, icon: Icon, name?: string) => {
```
KaWaite, line 77: Move this function out of the main component so it renders only once and cleans things up a bit

ui/src/features/Editor/components/LeftPanel/index.tsx
```diff
@@ -28,46 +40,112 @@ const LeftPanel: React.FC<Props> = ({
...
+
+  useEffect(() => {
+    if (selected && nodeId) {
+      setPreviousZoom(getZoom());
+    }
+
+    if (!isOpen) {
+      setNodeId(undefined);
+    }
+  }, [selected, isOpen, getZoom, nodeId]);
```
KaWaite, line 69: I'm not sure this is working well. It zooms to the node, but doesn't return to the previous zoom. (I think actually there is no way to deselect, so that might be why).

Also, is there any way you can move this logic in to a callback that handles selection or something? Using useEffects in this way often causes headaches later.

ui/src/features/Editor/components/LeftPanel/index.tsx
```diff
@@ -28,46 +40,112 @@ const LeftPanel: React.FC<Props> = ({
...
+  // onNodeDoubleClick,
+  selected,
 }) => {
   const t = useT();
   const { workspaceId } = useParams({ strict: false });
   const [selectedTab, setSelectedTab] = useState<Tab | undefined>();
+  const { fitView, getZoom, zoomTo } = useReactFlow();
+  const [previousZoom, setPreviousZoom] = useState<number | undefined>(
+    undefined,
+  );
```
KaWaite, line 52: I think previousZoom can be a ref. Don't think we need/want to rerender on this

ui/src/features/Editor/components/LeftPanel/index.tsx
KaWaite, line 77: If you add nodes to the params, you can move this function out of the main component so it renders only once.

KaWaite (approved):

PR #877, opened 2025-02-18
fix(ui): added temp fix for note text

billcookie:
# Overview
Note text from textarea was not saving correctly. 

## What I've done
Temp fix is adding text via params


ui/src/features/Editor/components/Canvas/components/Nodes/NoteNode/index.tsx
```diff
@@ -108,7 +110,23 @@ const NoteNode: React.FC<NoteNodeProps> = ({ data, ...props }) => {
           <Note />
           <p>{data.params?.customName ?? data.officialName}</p>
         </div>
-        <textarea
+        <span
```
KaWaite, line 113: If there is no real reason to make this a span, can you do a div (bg color) + p (obv text stuff) instead? 🙏🏻

KaWaite (approved):

PR #891, opened 2025-02-20
fix(ui): Text Area and Input Fix for Params

billcookie:
# Overview
As we changed the RJSF BaseInputTemplate to be a textarea rather than an input it caused all params to be a text area (e.g. color picker etc)

## What I've done
Params that have a different format like color will now be input with a simple conditional. This might be expanded in the future as we expand param customization etc.
## How I tested
Manually
## Summary by CodeRabbit

- **New Features**
	- Form inputs now adapt based on the field type, allowing for specialized controls for color inputs.
  
- **Refactor**
	- Note display has been streamlined by removing redundant background styling and outdated text input elements, resulting in a cleaner and more consistent interface.

### DIFF
--- a/ui/src/components/SchemaForm/Templates/BaseInputTemplate.tsx
+++ b/ui/src/components/SchemaForm/Templates/BaseInputTemplate.tsx
@@ -77,7 +77,7 @@ const BaseInputTemplate = <
       onChangeOverride || onChange(value === "" ? options.emptyValue : value)
     );
   };
-  // For most text-based params we want TextArea. But for certain schema format types, we want Input to get the appropriate styling @billcookie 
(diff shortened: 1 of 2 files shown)


ui/src/components/SchemaForm/Templates/BaseInputTemplate.tsx
```diff
@@ -77,7 +77,27 @@ const BaseInputTemplate = <
       onChangeOverride || onChange(value === "" ? options.emptyValue : value)
     );
   };
-
+  // We want the default to be textarea as for most params they will be long strings, however, for color and future formats we can set as input to get the correct styling @billcookie
```
KaWaite, line 80:
```suggestion
  // For most text-based params we want TextArea. But for certain schema format types, we want Input to get the appropriate styling @billcookie 
```

KaWaite (approved):

PR #894, opened 2025-02-21
fix(ui): Added dialog for adding members to a team workspace

billcookie:
# Overview
Currently, there is a text area to enter an email that looks like what you'd use to search existing members, and then click "Add Member" to add.

Its confusing, and what should be is clicking "Add Member" should open a dialog where a user can be added. 

## What I've done
Added a dialog for when a user wants to add a new member to a team workspace and the button is also hidden for personal workspaces.
## How I tested
Manually


ui/src/features/WorkspaceSettings/components/MembersSettings/components/MemberAddDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <DialogTitle>{t("Add a New User")}</DialogTitle>
+        <DialogContentWrapper>
+          <DialogContentSection className="flex flex-col">
+            <Input
+              className="w-full"
+              placeholder={t("Enter email")}
+              value={email}
+              onChange={(e) => setEmail(e.target.value)}
+            />
+            <p className="text-sm text-red-400">{error}</p>
```
KaWaite, line 44: This still remains on the original component too. Please remove it from there since its here now

ui/src/features/WorkspaceSettings/components/MembersSettings/index.tsx
```diff
@@ -69,35 +73,48 @@ const MembersSettings: React.FC = () => {
...
-    setError(undefined);
     if (!currentWorkspace?.id) return;
     const { workspace } = await updateMemberOfWorkspace(
       currentWorkspace.id,
       userId,
       role,
     );
     if (!workspace) {
-      setError(t("Failed to change role of the member"));
+      toast({
```
KaWaite, line 95: You have toasts in useApi as well as here now. Currently we are mostly handling toasts around gql inside of useApi, so remove here please.

KaWaite (approved):

PR #947, opened 2025-03-02
feat(engine): Implemented LineOnLineOverlayer

TadaTeruki:
# Overview
- Implemented `LineOnLineOverlayer` as a new action.
- Added `line_ops` and `line_string_ops` package to define additional operations on a line and a line string.

## How I tested

```
$ cd engine
$ yaml-include runtime/examples/fixture/workflow/requirement/plateau4/a001-7/line_on_line_overlayer.yml | cargo run --package reearth-flow-cli -- run --var="outputPath=${output path}" --workflow -
```

### Input
Screenshot from 2025-03-02 17-49-59

### Result (Line)
Screenshot from 2025-03-02 17-49-47

### Result (Point)
Screenshot from 2025-03-02 17-50-15

## Memo

- I implemented `LineOps::difference()` in `engine/runtime/geometry/src/algorithm/line_ops/difference.rs` to remove overlapping parts of lines, but it is currently unused and can be removed.
- `LineString2D` doesn't implement `LineStringOps` because it has an $O(N^2)$ time complexity. Instead, I implemented `LineStringWithTree2D`, which uses an R-tree to manage its lines.
- parallelized using rayon

## Summary by CodeRabbit

- **New Features**
  - Introduced a new workflow configuration for processing GeoJSON data with line overlay operations.
  - Enhanced geometry capabilities with advanced operations for calculating intersections, differences, and splits between line features.
  - Added a feature to compute the length of line segments.

- **Refactor**
  - Improved the overlay processing logic for better efficiency, including support for parallel processing.

- **Chore**
  - Updated processor configuration schemas by replacing the output attribute with a tolerance parameter.


miseyu (approved):
Thank you so much!
LGTM

PR #973, opened 2025-03-05
feat(ui): subscription set up for real-time logs

billcookie:
# Overview

## Summary by CodeRabbit

- **New Features**
  - Enabled real-time log monitoring with live updates for active jobs.
  - Added a “Node Id” column in log displays to provide enhanced context.
  - Expanded localization support to include the new UI term across multiple languages.

- **Refactor**
  - Improved log display consistency with standardized timestamps and log level formatting.
  - Enhanced job detail views to offer clearer status feedback during project execution.


ui/src/lib/gql/subscriptions/index.ts
```diff
@@ -0,0 +1,25 @@
+import { createClient } from "graphql-ws";
+import { useMemo } from "react";
+
+import { config } from "@flow/config";
+import { useAuth } from "@flow/lib/auth";
+
+export const useWsClient = () => {
```
KaWaite, line 7: I think this should be setup as a provider like the other gql providers and wrap the app. If not, when we have multiple subscriptions on a single page we will be setting up multiple clients...

KaWaite (approved):

PR #1045, opened 2025-03-15
feat(engine): Ensure cleanup call is executed in both success and failure cases in command.rs [FLOW-BE-94]

soma00333:
# Overview
This pull request includes changes to the `engine/worker/src/command.rs` file to adjust the cleanup process in the `RunWorkerCommand` implementation. The most important changes involve moving the cleanup call to ensure it is always executed regardless of the job result.

Changes to cleanup process:

* [`engine/worker/src/command.rs`](diffhunk://#diff-b76379618575e6337dfe5e8e3653b74bde84b2d8a8ad9836ae52da1afdf7fac9L174): Removed the call to `self.cleanup(&meta, &storage_resolver).await?` from within the success condition of the job result match block.
* [`engine/worker/src/command.rs`](diffhunk://#diff-b76379618575e6337dfe5e8e3653b74bde84b2d8a8ad9836ae52da1afdf7fac9R182): Added the call to `self.cleanup(&meta, &storage_resolver).await?` after determining the job result, ensuring it runs regardless of the job outcome.
## Summary by CodeRabbit

- **Bug Fixes**
	- Updated system behavior to ensure that cleanup tasks are executed following every job process, regardless of outcome. This change promotes smoother operation, enhanced stability, and reduces issues that could arise from conditional cleanup execution, contributing to a more consistent overall experience.

### DIFF
--- a/engine/worker/src/command.rs
+++ b/engine/worker/src/command.rs
@@ -171,7 +171,6 @@ impl RunWorkerCommand {
         let job_result = match result {
             Ok(_) => {
                 if node_failure_handler.all_success() {
-                    self.cleanup(&meta, &storage_resolver).await?;
                     JobResult::Success
                 } else {
                     tracing::error!("Failed nodes: {:?}", node_failure_handler.failed_nodes());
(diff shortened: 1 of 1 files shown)


miseyu (approved):
LGTM

PR #1127, opened 2025-03-27
fix(ui): added selection context menu and deletion

billcookie:
# Overview

## Summary by CodeRabbit

- **New Features**
  - Introduced distinct context menus for individual nodes versus multiple selections.
  - Added a dedicated selection menu option that facilitates deletion of selected items.
  - Enhanced canvas and editor interactions with improved edge selection support.
  - Optimized context menu positioning for a more intuitive user experience.
  - Expanded localization by adding the “Delete Selection” option across multiple languages.

### DIFF
--- a/ui/src/components/ContextMenu/index.tsx
+++ b/ui/src/components/ContextMenu/index.tsx
@@ -19,7 +19,7 @@ export type ContextMenuMeta =
   | (ContextMenuStyles & SelectionContextMenuMeta);
 
 type ContextMenuProps = {
-  items: ContextMenuItemProps[];
+  items: ContextMenuItemType[];
   contextMenuMeta: ContextMenuMeta;
 };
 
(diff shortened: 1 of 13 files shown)


ui/src/components/ContextMenu/index.tsx
```diff
@@ -1,10 +1,23 @@
...
 
+type NodeContextMenuMeta = {
+  type: "node";
+  data: Node;
+};
+
+type SelectionContextMenuMeta = {
+  type: "selection";
+  data: Node[];
+};
```
KaWaite, line 15: is `type: "node" | "nodes"` being used at all inside the component? It doesn't look like it, so if not you can just do something like this instead:
```
type ContextMenuMeta<T = any> = {
  data: T;
}
```
Then use the ContextMenu like 
```
<ContextMenu<Node[]> items={menuItems} contextMenuMeta={contextMenu} />;
```

KaWaite (approved):

PR #1166, opened 2025-04-17
fix(ui): workflow tab name update

billcookie:
# Overview

## Summary by CodeRabbit

- **New Features**
  - Workflow renaming is now triggered from the parameters panel for subworkflow nodes when customizations are updated.

- **Refactor**
  - Workflow renaming logic and related callbacks were removed from the top bar and workflow tabs, streamlining the interface and simplifying workflow tab interactions.
  - Simplified internal workflow renaming logic to directly update workflow names.
  - Updated workflow names state management for improved synchronization.
  - Shifted workflow rename event handling from the top bar to the parameters panel within the editor.

- **Style**
  - Removed unused translation strings related to workflow renaming errors from all supported languages.

### DIFF
--- a/ui/src/features/Editor/components/ParamsPanel/components/ParamEditor/index.tsx
+++ b/ui/src/features/Editor/components/ParamsPanel/components/ParamEditor/index.tsx
@@ -30,7 +30,7 @@ type Props = {
     data: any,
     type: "params" | "customizations",
   ) => Promise<void>;
-  onWorkflowRename: (id: string, name: string) => void;
+  onWorkflowRename?: (id: string, name: string) => void;
 };
 
 const ParamEditor: React.FC<Props> = ({
@@ -83,7 +83,7 @@ const ParamEditor: React.FC<Props> = ({
       onUpdate(nodeId, updatedParams, "params");
     } else if (nodeType === "subworkflow" && nodeMeta.subworkflowId) {
       onUpdate(nodeId, updatedCustomization, "customizations");
-      onWorkflowRename(
+      onWorkflowRename?.(
         nodeMeta?.subworkflowId,
         updatedCustomization?.customName || nodeMeta?.officialName,
       );
--- a/ui/src/features/Editor/components/ParamsPanel/components/ParamEditor/index.tsx
+++ b/ui/src/features/Editor/components/ParamsPanel/components/ParamEditor/index.tsx
@@ -30,6 +30,7 @@ type Props = {
     data: any,
     type: "params" | "customizations",
   ) => Promise<void>;
+  onWorkflowRename: (id: string, name: string) => void;
 };
 
 const ParamEditor: React.FC<Props> = ({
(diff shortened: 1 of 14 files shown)


ui/src/lib/yjs/useWorkflowTabs.ts
```diff
@@ -27,7 +32,7 @@ export default ({
         id: w2.id as string,
         name: w2.name as string,
       })),
-    [rawWorkflows.length], // eslint-disable-line react-hooks/exhaustive-deps
+    [rawWorkflows.length, workflowName], // eslint-disable-line react-hooks/exhaustive-deps
```
KaWaite, line 35: Due to workflowName having rawworkflows as a dependency, this means when rawWorkflows changes (with each change), workflowName will update, which means workflows will update. So same behaviour we are tring to avoid with the rawWorkflows.length.

ui/src/features/Editor/components/ParamsPanel/components/ParamEditor/index.tsx
```diff
@@ -79,6 +81,12 @@ const ParamEditor: React.FC<Props> = ({
   const handleUpdate = () => {
     if (activeTab === "params") {
       onUpdate(nodeId, updatedParams, "params");
+    } else if (nodeType === "subworkflow" && nodeMeta.subworkflowId) {
+      onUpdate(nodeId, updatedCustomization, "customizations");
+      onWorkflowRename?.(
+        nodeMeta?.subworkflowId,
+        updatedCustomization?.customName || nodeMeta?.officialName,
+      );
```
KaWaite, line 89: Need to lint this probably.
```suggestion
    } else if (nodeType === "subworkflow" && nodeMeta.subworkflowId) {
      onUpdate(nodeId, updatedCustomization, "customizations");
      if (updatedCustomization?.customName) {
        onWorkflowRename?.(
          nodeMeta?.subworkflowId,
          updatedCustomization.customName,
        );
      }
```

billcookie: I think for this the original is probably ok as if the user has not added a customName and has left it blank we would want the official name to be used (for other nodes this is the case).

ui/src/lib/yjs/useWorkflowTabs.ts
```diff
@@ -18,17 +18,24 @@ export default ({
...
-        id: w2.id as string,
-        name: w2.name as string,
-      })),
-    [rawWorkflows.length], // eslint-disable-line react-hooks/exhaustive-deps
-  );
+  useEffect(() => {
+    if (rawWorkflows.length !== workflowNames.length) {
+      setWorkflowsNames(rawWorkflows.map((w) => ({ id: w.id, name: w.name })));
+    }
+  }, [rawWorkflows.length, workflowNames.length]); // eslint-disable-line react-hooks/exhaustive-deps
```
KaWaite, line 31: I feel like its better to keep updating consistent, which means handling workflow add/removal should happen in the same way as the rename in the Editor hooks.

KaWaite (approved):

PR #1191, opened 2025-05-21
fix(ui): removed console.log and set edges to deselect on paste [FLOW-FE-167]

billcookie:
# Overview

## What I've done
Previously selected edges from the copied/cut nodes were not being deselected, therefore added similar logic of hanldeNodesChange for handleEdgesChange

### DIFF
--- a/ui/src/features/Editor/useCanvasCopyPaste.ts
+++ b/ui/src/features/Editor/useCanvasCopyPaste.ts
@@ -395,8 +395,8 @@ export default ({
         selected: false,
       }));
 
-      const edgeChanges: EdgeChange[] = edges.map((n) => ({
-        id: n.id,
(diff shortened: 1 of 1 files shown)


ui/src/features/Editor/useCanvasCopyPaste.ts
```diff
@@ -396,7 +395,14 @@ export default ({
         selected: false,
       }));
 
+      const edgeChanges: EdgeChange[] = edges.map((n) => ({
+        id: n.id,
+        type: "select",
+        selected: false,
+      }));
```
KaWaite, line 402:
```suggestion
      const edgeChanges: EdgeChange[] = edges.map((e) => ({
        id: e.id,
        type: "select",
        selected: false,
      }));
```

KaWaite (approved):

PR #1204, opened 2025-05-26
fix(ui): prevent deletion in shared canvas

billcookie:
# Overview

### DIFF
--- a/ui/src/features/Canvas/components/CanvasContextMenu/index.tsx
+++ b/ui/src/features/Canvas/components/CanvasContextMenu/index.tsx
@@ -52,7 +52,7 @@ const CanvasContextMenu: React.FC<Props> = ({
   const nodes = Array.isArray(data) ? data : undefined;
   const node = Array.isArray(data) ? undefined : data;
 
-  const handleSecondaryNodeAction = useCallback(
+  const handleNodeSettingsOpen = useCallback(
     (node: Node) => {
       onNodeSettings?.(undefined, node.id);
     },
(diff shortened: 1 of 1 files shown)


ui/src/features/Canvas/index.tsx
```diff
@@ -32,6 +32,7 @@ const snapGrid: SnapGrid = [gridSize, gridSize];
 const proOptions: ProOptions = { hideAttribution: true };
 
 type Props = {
+  isSharedCanvas: boolean;
```
KaWaite, line 35:
```suggestion
  isSharedCanvas?: boolean;
```
If you make this optional, you don't need to pass anything when it is not a shared canvas, and then you can reduce `isSharedCanvas={true}` to just `isSharedCanvas`. Keeps things cleaner when using `Canvas`

ui/src/features/Editor/index.tsx
```diff
@@ -107,6 +107,7 @@ export default function Editor({
               onWorkflowRedo={handleWorkflowRedo}
               onLayoutChange={handleLayoutChange}>
               <Canvas
+                isSharedCanvas={false}
```
KaWaite, line 110: If you make the change in Canvas, you can remove here

ui/src/features/SharedCanvas/index.tsx
```diff
@@ -58,6 +58,7 @@ const SharedCanvas: React.FC<Props> = ({
     <div className="relative flex size-full flex-col">
       <EditorProvider value={editorContext}>
         <Canvas
+          isSharedCanvas={true}
```
KaWaite, line 61: If you make the change in Canvas,
```suggestion
          isSharedCanvas
```

ui/src/features/Canvas/components/CanvasContextMenu/index.tsx
```diff
@@ -104,7 +100,7 @@ const CanvasContextMenu: React.FC<Props> = ({
           shortcut: (
             <ContextMenuShortcut keyBinding={{ key: "c", commandKey: true }} />
           ),
-          disabled: !nodes && !node,
+          disabled: (!nodes && !node) || !onCut,
```
KaWaite, line 103:
```suggestion
          disabled: (!nodes && !node) || !onCopy,
```

KaWaite (approved):

PR #1206, opened 2025-05-28
fix(ui): Read only and onClose params fix for Shared Canvas [FLOW-FE-176]

billcookie:
# Overview


ui/src/components/SchemaForm/index.tsx
```diff
@@ -16,13 +16,15 @@ type SchemaFormProps = {
   defaultFormData?: any;
   onChange: (data: any) => void;
   onError?: (errors: any[]) => void;
+  disableEditing?: boolean; // Optional prop to disable editing
```
KaWaite, line 19: `readOnly`

ui/src/features/Editor/components/ParamsPanel/components/ParamEditor/index.tsx
```diff
@@ -31,6 +31,7 @@ type Props = {
     type: "params" | "customizations",
   ) => Promise<void>;
   onWorkflowRename?: (id: string, name: string) => void;
+  disableEditing?: boolean;
```
KaWaite, line 34: `readOnly`

ui/src/features/Editor/components/ParamsPanel/components/ParamEditor/index.tsx
```diff
@@ -40,6 +41,7 @@ const ParamEditor: React.FC<Props> = ({
   // nodeParameters = [{ id: "param1", name: "Param 1", value: "Value 1", type: "string"}],
   onUpdate,
   onWorkflowRename,
+  disableEditing,
```
KaWaite, line 44: `readOnly`

ui/src/components/SchemaForm/Templates/BaseInputTemplate.tsx
```diff
@@ -143,6 +143,7 @@ const BaseInputTemplate = <
               variant="outline"
               size="sm"
               onClick={() => onChange(defaultValue)}
+              disabled={disabled || readonly}
```
KaWaite, line 146:
```suggestion
              disabled={readonly || disabled}
```
I think readOnly should be prioritized. In a readOnly state, everything should definitively be disabled no matter what

ui/src/components/SchemaForm/index.tsx
```diff
@@ -12,13 +12,15 @@ import { useT } from "@flow/lib/i18n";
 import { ThemedForm } from "./ThemedForm";
 
 type SchemaFormProps = {
+  readonly?: boolean; // Optional prop to disable editing
```
KaWaite, line 15: Comment doesn't really add anything. Please remove it

ui/src/components/SchemaForm/index.tsx
```diff
@@ -12,13 +12,15 @@ import { useT } from "@flow/lib/i18n";
...
   schema?: RJSFSchema;
   defaultFormData?: any;
   onChange: (data: any) => void;
   onError?: (errors: any[]) => void;
 };
 
 const SchemaForm: React.FC<SchemaFormProps> = ({
+  readonly = false,
```
KaWaite, line 23: Since readonly is optional, we shouldn't need to set a default value here.

ui/src/features/Canvas/index.tsx
```diff
@@ -114,20 +114,20 @@ const Canvas: React.FC<Props> = ({
       // fitView
       ref={paneRef}
       // Locking props START
-      nodesDraggable={!canvasLock}
-      nodesConnectable={!canvasLock}
-      nodesFocusable={!canvasLock}
-      edgesFocusable={!canvasLock}
+      // nodesDraggable={!canvasLock}
+      nodesConnectable={!readonly}
```
KaWaite, line 118: Sorry, I know this goes against what I said, but if we are going to set nodesConnectable based on a readonly state, it logically makes sense (and is clearer to a code reader) to disable/enable everything together, deliberately.

ui/src/features/Editor/index.tsx
```diff
@@ -135,11 +134,14 @@ export default function Editor({
             project={currentProject}
             yDoc={yDoc}
           />
-          <ParamsPanel
-            selected={locallyLockedNode}
-            onDataSubmit={handleNodeDataUpdate}
-            onWorkflowRename={handleWorkflowRename}
-          />
+          {openNode && (
```
KaWaite, line 137: Since we are relying on ParamsPanel's dialog to open/close it, we shouldn't need to conditionally render the ParamsPanel here with openNode &&

ui/src/features/Canvas/index.tsx
billcookie: Hmmm I understand what you mean, however, this is the way to prevent handles from being used on nodes without excessive prop drilling or a larger refactor of how CustomHandle components are being used (which I feel would be unnecessary for such a small benefit). 

In that sense what params should be read only and not? I am unsure. Surely we would want zoomOnPinch, zoomOnDoubleClick etc to be not read only. Just basing this on what we had previously on !canvaslock comment.

ui/src/features/Editor/components/ParamsPanel/index.tsx
```diff
@@ -60,26 +57,27 @@ const ParamsPanel: React.FC<Props> = ({
...
 
   return (
-    <Dialog open={!!selected} onOpenChange={handleClose}>
+    <Dialog open={!!openNode} onOpenChange={handleClose}>
       <DialogContent size="2xl">
         <DialogHeader>
           <DialogTitle>{t("Parameter Editor")}</DialogTitle>
         </DialogHeader>
-        {selected && (
+        {openNode && (
```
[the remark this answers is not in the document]
KaWaite: @billcookie This one is unaddressed

KaWaite (approved):

KaWaite (commented):
Just un-approving it

KaWaite (changes requested):
asdf

ui/src/features/Editor/components/ParamsPanel/index.tsx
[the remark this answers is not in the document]
billcookie: Little confused about this as this code was the same as before this PR apart from a name change. We need nodeId to be present for the component to load. 🤔

KaWaite (approved):

PR #1208, opened 2025-05-30
fix(ui): Import via Shared Url to Workspaces [FLOW-FE-175] 

billcookie:
# Overview


ui/src/features/Editor/components/TopBar/components/ActionBar/index.tsx
```diff
@@ -94,7 +100,7 @@ const ActionBar: React.FC<Props> = ({
               alignOffset={2}>
               <DropdownMenuItem
                 className="flex justify-between gap-4"
-                onClick={handleProjectExport}>
+                onClick={() => handleProjectExport({ yDoc, project })}>
```
KaWaite, line 103: Rather than passing yDoc and project in so that you need to do this kind of calling of handleProjectExport, its cleaner and more succinct I think to just raise handleProjectExport's use up, preferably where the other callback props are originating (or close to it)

ui/src/lib/gql/user/useApi.ts
```diff
@@ -20,6 +31,30 @@ export const useUser = () => {
...
+          role: member.role,
+          user: member.user
+            ? {
+                id: member.user.id,
+                email: member.user.email,
+                name: member.user.name,
+              }
+            : undefined,
+        })),
+      })),
```
KaWaite, line 53: I think you copy/pasted this from `createNewWorkspaceObject` found in the `workspace/useQueries.ts` file. Instead of just copy/pasting, please move the original function into gql/convert, and use that here and elsewhere.

KaWaite (approved):

PR #1221, opened 2025-06-13
feat(ui): manual save via button or shortcut [FLOW-FE-153]

billcookie:
# Overview
A new feature that adds manual saving via a button or cmd + s


ui/src/lib/gql/document/useQueries.ts
```diff
@@ -68,6 +68,25 @@ export const useQueries = () => {
       refetchOnWindowFocus: false,
     });
 
+  const usePreviewSnapshot = useMutation({
```
KaWaite, line 71: Yet again I'm getting mixed up. How is this different from useProjectSnapshotQuery? It being a mutation is confusing.
Whichever it is, the name should meet the appropriate naming convention

KaWaite (approved):

PR #1229, opened 2025-06-16
chore(ui): eslint-plugin-better-tailwindcss setup [FLOW-FE-148]

billcookie:
# Overview
Original Tailwind linter we had been using has not been updated to V4 and [does not seem it will likely be anytime soon](https://github.com/francoismassart/eslint-plugin-tailwindcss/issues/325). Therefore, have chosen to [test this linter](https://github.com/schoero/eslint-plugin-better-tailwindcss).
## What I've done
Setup new linter
## Memo
It is to be noted that some rules that we had on the old linter are not available. They have been kept commented out incase they become available or the old linter we used is updated.

### DIFF
--- a/ui/eslint.config.mjs
+++ b/ui/eslint.config.mjs
@@ -9,28 +9,6 @@ const storyBookConfig = {
     storybook,
   },
 };
-// eslint-plugin-tailwindcss does not support the latest tailwindcss version.
-// Therefore, we use eslint-plugin-better-tailwindcss instead, which is being regularly updated and a work in progress.
-// const customTailwindConfig = [
-//   ...tailwind.configs["flat/recommended"],
-//   {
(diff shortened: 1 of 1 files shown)


ui/eslint.config.mjs
```diff
@@ -8,7 +9,8 @@ const storyBookConfig = {
     storybook,
   },
 };
-
+// eslint-plugin-tailwindcss does not support the latest tailwindcss version.
+// Therefore, we use eslint-plugin-better-tailwindcss instead, which is being regularly updated and a work in progress.
```
KaWaite, line 13: This kinda detail isn't necessary. The original plugin is irrelevant for us and the repo now.

KaWaite (dismissed):

PR #1240, opened 2025-06-19
fix(ui): Map Libre Improvements for All Geometry types [FLOW-FE-213]

billcookie:
# Overview
This PR improves map visualization by replacing manual bounds calculation with the Turf.js bbox function and adding interactive layers for selecting features.

- Replaces manual GeoJSON bounds computation with Turf's bbox.
- Adds interactiveLayerIds and onClick handler to update selected features.
- Updates package.json to include the @turf/bbox dependency.


ui/src/components/visualizations/MapLibre/index.tsx
```diff
@@ -80,16 +69,25 @@ const MapLibre: React.FC<Props> = ({ className, fileContent, fileType }) => {
...
-        });
+      try {
+        if (
+          !selectedFeature.geometry ||
+          !selectedFeature.geometry.type ||
+          !selectedFeature.geometry.coordinates
+        ) {
+          console.error("Invalid selectedFeature geometry:", selectedFeature);
+          return;
+        }
```
KaWaite, line 80: Isn't this kind of redundant?  If bbox throws an error due to selectedFeature not being adequate shouldn't the try catch catch the error? You also aren't doing this elsewhere when using bbox.

PR #1241, opened 2025-06-19
fix(ui): added a generic error page for general errors [FLOW-FE-212]

billcookie:
# Overview
For rendering errors there was no generic message or indication to the user there is a problem. Also for any of these types of errors that occurred in the Editor it would prompt the user to revert project when it was unnecessary.

Adds a reusable generic error page and updates routes to distinguish corrupted-project errors from other failures.

- Introduces ErrorPage component under components/errors and exports it.
- Custom Errors for edge cases (ProjectCorruption)
- Fallback at root level to 404 if page route doesnt exist
- Fixed logo being a green square

## What I've done
Added a check for when a project is corrupted by checking the instance of the error to see if it matches the new ProjectCorruptionError if it does we can conclude it is the project corruption error where a position is not found. If this error is not present then we just render a something went wrong page similar to that of the "user not found".

This has all been combined to be in a reusable component and has been added to components under `errors` directory.

### DIFF
--- a/ui/src/routes/index.tsx
+++ b/ui/src/routes/index.tsx
@@ -1,9 +1,11 @@
 import { createFileRoute, redirect } from "@tanstack/react-router";
 
 import { LoadingSplashscreen } from "@flow/components";
+import ErrorPage from "@flow/components/errors/ErrorPage";
 
 export const Route = createFileRoute("/")({
   component: () => <LoadingSplashscreen />,
+  errorComponent: () => <ErrorPage errorMessage={"Something Went Wrong"} />,
   loader: () => {
     throw redirect({ to: "/workspaces" });
   },
--- a/ui/src/routes/index.tsx
+++ b/ui/src/routes/index.tsx
@@ -5,7 +5,7 @@ import ErrorPage from "@flow/components/errors/ErrorPage";
 
 export const Route = createFileRoute("/")({
   component: () => <LoadingSplashscreen />,
-  errorComponent: () => <ErrorPage errorMessage={"Something Went Wrong"} />,
+  errorComponent: () => <ErrorPage />,
   loader: () => {
     throw redirect({ to: "/workspaces" });
   },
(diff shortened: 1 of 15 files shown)


ui/src/routes/workspaces.tsx
```diff
@@ -21,6 +23,8 @@ export const Route = createFileRoute("/workspaces")({
       <WorkspaceRoute />
     </AuthenticationWrapper>
   ),
+  errorComponent: () => <ErrorPage errorMessage={"Something Went Wrong"} />,
```
KaWaite, line 26: Error message should be internationalized

ui/src/routes/index.tsx
```diff
@@ -1,9 +1,11 @@
 import { createFileRoute, redirect } from "@tanstack/react-router";
 
 import { LoadingSplashscreen } from "@flow/components";
+import ErrorPage from "@flow/components/errors/ErrorPage";
 
 export const Route = createFileRoute("/")({
   component: () => <LoadingSplashscreen />,
+  errorComponent: () => <ErrorPage errorMessage={"Something Went Wrong"} />,
```
KaWaite, line 8: Same. Internationalization

ui/src/routes/shared.$sharedToken.lazy.tsx
```diff
@@ -92,15 +95,20 @@ const EditorComponent = ({ accessToken }: { accessToken?: string }) => {
...
 
   const { yWorkflows, yDocState, isSynced, undoTrackerActionWrapper } =
     useYjsSetup({
       projectId: sharedProject?.id,
       workflowId: DEFAULT_ENTRY_GRAPH_ID,
     });
 
-  return !yWorkflows || !isSynced || !undoTrackerActionWrapper ? (
+  return isError ? (
+    <NotFound />
```
KaWaite, line 107: NotFound doesn't really apply to isError in useGetSharedProject, does it? NotFound should be only used if a route is wrong as a fallback. Not for errors.

KaWaite (approved):

PR #1250, opened 2025-06-26
fix(ui): Reviewed Toasts, API fallbacks,  Reusable RenderFallback Component [FLOW-FE-220]

billcookie:
# Overview
This PR checks how we are currently handling success, errors, fallbacks and attempts to unify and make them better.

## What I've done

- Added Reusable RenderFallback Component to allow UI fallback if rendering issues instead of crashing the app.
- Added toasts were appropriate
- Reviewed current API fallbacks (i.e data undefined)


ui/src/features/Editor/components/OverlayUI/components/DebugPreview/index.tsx
```diff
@@ -134,17 +135,35 @@ const DebugPreview: React.FC = () => {
...
             className="m-0 h-[calc(100%-32px)] p-1"
             value="3d-viewer">
-            <GeoMap fileContent={selectedOutputData} fileType={fileType} />
+            <RenderFallback
+              message={t(
+                "3D Viewer Could Not Be Loaded. Check if the data is valid.",
+              )}
+              textSize="sm">
+              <GeoMap fileContent={selectedOutputData} fileType={fileType} />
+            </RenderFallback>
```
KaWaite, line 166: I'd prefer RenderFallback to do more. That way we could place it at a higher position and reduce code in components that don't care about the error message, etc.

billcookie: If we place in a higher position such as the tab level then an error will occur. For example, if MapLibre has issues then all components will share the error message and will not render. I am a little unclear in what you mean by "do more"? 🤔. My understanding for error boundaries is that they should render some form of UI if there is an error in areas we believe could be error prone. This could either be route level or something more specific like map viewers.

KaWaite (approved):

PR #1256, opened 2025-06-30
feat(ui): graphql and mock data setup for assets

billcookie:
# Overview


ui/src/components/DataTable/index.tsx
```diff
@@ -138,12 +148,31 @@ function DataTable<TData, TValue>({
...
-              value={globalFilter ?? ""}
-              onChange={(e) => setGlobalFilter(String(e.target.value))}
+              value={searchTerm ?? globalFilter}
+              onChange={(e) => {
+                const value = String(e.target.value);
+                if (setSearchTerm) {
+                  setSearchTerm(value);
+                }
+                setGlobalFilter(value);
+              }}
```
KaWaite, line 158: I feel like search term and global filter shouldn't be mixed. In my mind they should be able to work together (unless I'm misunderstanding what global filter would be). 
Just like with the gql having searchTerm and order, etc separated, is there any way here you can keep them working independently? 🙏🏻

ui/src/features/Editor/components/TopBar/components/AssetsDialog/AssetDeletionDialog.tsx
```diff
@@ -0,0 +1,30 @@
...
+  assetToBeDeleted,
+  setAssetToBeDeleted,
+  onDeleteAsset,
+}) => {
+  const t = useT();
+  return (
+    <ConfirmationDialog
+      title={t("Are you absolutely sure?")}
+      description={t(
+        "This action cannot be undone. This will permanently delete your asset. Deletion of an asset could cause issues projects or deployments that are referencing this asset.",
```
KaWaite, line 20: Please double check the language is clear enough

ui/src/global-constants.ts
```diff
@@ -4,4 +4,6 @@ export const DEFAULT_ROUTING_PORT = "default";
 export const DEFAULT_NODE_SIZE = { width: 150, height: 25 };
 export const ALLOWED_WORKFLOW_FILE_EXTENSIONS = ".json, .yaml, .yml";
 export const ALLOWED_PROJECT_IMPORT_EXTENSIONS = ".zip";
+export const ALLOWED_ASSET_IMPORT_EXTENSIONS =
+  ".csv, .geojson, .gml, .json, .sql, .tsv";
```
KaWaite, line 8: .sql doesn't exist. Sorry when I explained the allowed asset imports I should have caught that. We don't need to do anything with SQL since that functionality is fully action/engine related.

ui/src/features/Editor/components/TopBar/hooks.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export type DialogOptions =
+  | "deploy"
+  | "share"
+  | "version"
+  | "assets"
+  | "debugStop"
+  | "projectVariables"
+  | undefined;
+export default () => {
```
KaWaite, line 18:
```suggestion
  | undefined;
  
export default () => {
```

ui/src/features/Editor/components/TopBar/components/AssetsDialog/AssetDeletionDialog.tsx
KaWaite: Not sure if the text here was changed, but the grammar seems off.
Please use:

"This action cannot be undone. Deleting this asset will permanently remove it from your account and may cause issues with any projects or deployments that reference it. Please ensure this asset is no longer in use before proceeding."

PR #1269, opened 2025-07-07
fix(ui): refactor for combined near full screen for debug panel [FLOW-FE-209]

billcookie:
# Overview

This PR changes the way we display the debug panels (`DebugLogs` and `DebugPreview`) for when a workflow debug run has finished.  
## What I've done

- Created a new component and refactored logic into `DebugPanel`.
- Allowed users to view 2D and 3D viewers in full screen mode
- Added resizer to allow users to adjust the panels to their needs

## Screenshot
Previous: 

Screenshot 2025-07-07 at 15 28 47

Current (WIP): 
Screenshot 2025-07-07 at 15 30 49


ui/src/components/visualizations/MapLibre/index.tsx
```diff
@@ -89,15 +89,15 @@ const MapLibre: React.FC<Props> = ({ className, fileContent, fileType }) => {
...
           />
         )}
+        {/* <FullscreenControl position="top-right" /> */}
+        {selectedFeature && (
+          <MapSidePanel
+            selectedFeature={selectedFeature}
+            setSelectedFeature={setSelectedFeature}
+            onFlyToSelectedFeature={handleFlyToSelectedFeature}
+          />
+        )}
```
KaWaite, line 99: If you are purposely putting this in the Map as a child, remove the parent div's relative class (also the div might not be needed anymore)

ui/src/components/visualizations/MapLibre/index.tsx
```diff
@@ -89,15 +89,15 @@ const MapLibre: React.FC<Props> = ({ className, fileContent, fileType }) => {
             onSelectedFeature={setSelectedFeature}
           />
         )}
+        {/* <FullscreenControl position="top-right" /> */}
```
KaWaite, line 92: Remove

ui/src/features/Editor/components/OverlayUI/components/DebugPanel/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+            </div>
+          )}
+        </div>
+      </div>
+      <TabsContent
+        value="debug-logs"
+        className="h-[calc(100%-35px)] overflow-scroll"
+        forceMount={debugJobIdRef.current !== debugJobId ? undefined : true}
+        hidden={tabValue !== "debug-logs"}>
+        <DebugLogs debugJobId={debugJobId} />{" "}
```
KaWaite, line 160:
```suggestion
        <DebugLogs debugJobId={debugJobId} />
```

ui/src/features/Editor/components/OverlayUI/components/DebugPanel/DebugPreview/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+}) => {
+  const t = useT();
+
+  return debugJobState && dataURLs ? (
+    <Tabs
+      className={`pointer-events-auto h-full w-full rounded-md bg-secondary p-1 shadow-md shadow-secondary transition-all`}
+      defaultValue="2d-viewer">
+      <div className="relative flex items-center">
+        <div className="flex w-full items-center p-1">
+          <TabsList className="flex w-full justify-between">
```
KaWaite, line 51: Please review the CSS. Divs with only one child don't need to be flexed or have items centered etc. Also a random relative class, etc. 
If you play around with CSS to get things working right, always do a post-step of removing stuff that seems not needed and reducing down to the least amount of CSS. Do the same for elements too.

ui/src/features/Editor/components/OverlayUI/components/DebugPanel/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+    if (dataURLs && !hasSwitchedToViewerRef.current) {
+      setTabValue("debug-viewer");
+      hasSwitchedToViewerRef.current = true;
+    }
+
+    if (!dataURLs) {
+      setTabValue("debug-logs");
+      hasSwitchedToViewerRef.current = false;
+    }
+  }, [dataURLs, debugJobId]);
```
KaWaite, line 79: Be very careful with doing too much in a single useEffect. Easier to read and safer typically to just have separate useEffects for each thing you want done.  In this case, if you split the first if statement into its own, we would easily be able to understand it'll only ever re-fire if debugJobId changes. Currently, we have to be cautious and think about undesired side-effects to debugJobIdRef if the dataURLs changes, something that is completely unrelated. Make sense?

ui/src/features/Editor/components/OverlayUI/components/DebugPanel/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+  return debugJobId ? (
+    <Tabs
+      className={`pointer-events-auto w-[95vw] overflow-hidden rounded-md bg-secondary shadow-md shadow-secondary transition-all ${minimized ? "h-[36px]" : fullscreenDebug ? "h-[100vh] w-[100vw]" : expanded ? "h-[80vh]" : "h-[500px]"}`}
+      value={tabValue}
+      defaultValue="debug-logs"
+      onDoubleClick={handleExpand}
+      onValueChange={setTabValue}>
+      <div className={`flex items-center ${minimized ? "" : "border-b"} p-1`}>
+        <div className="relative flex w-fit items-center p-1">
```
KaWaite, line 89: Here again a random relative with flex on a div that wraps only a single other div. Doesn't make sense

ui/src/features/Editor/components/OverlayUI/components/DebugPanel/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+      {dataURLs && (
+        <TabsContent
+          value="debug-viewer"
+          forceMount={true}
+          hidden={tabValue !== "debug-viewer"}
+          className="h-[calc(100%-35px)] overflow-scroll">
+          <ResizablePanelGroup className="h-full w-full" direction="horizontal">
+            <ResizablePanel defaultSize={70} minSize={20}>
+              <Tabs defaultValue="data-viewer">
+                <div className="relative flex w-fit items-center p-1">
```
KaWaite, line 171: Again relative and flex wrapping a single div (with no ascertainable absolute element

ui/src/features/Editor/components/OverlayUI/index.tsx
```diff
@@ -88,12 +92,15 @@ const OverlayUI: React.FC<OverlayUIProps> = ({
           </div>
         </div>
         <div id="right-top" className="absolute top-4 right-4" />
-        <div className="pointer-events-none absolute inset-y-2 bottom-4 left-4 flex items-end">
-          <DebugLogs />
+        <div
+          className={`pointer-events-none absolute inset-y-2 ${fullscreenDebug ? "bottom-0 left-0" : "bottom-4 left-4 "}   z-30 flex items-end`}>
```
KaWaite, line 96: If possible, I'd prefer keeping the absolute positioning static here, and then in the DebugPanel handle all DebugPanel specific positioning when in different states. Keeps the OverlayUI more simple while still allowing us to do unique things with the component itself.
Not super important though, so if you move the positioning to DebugPanel, negate the space in bottom and left, and it doesn't really work as expected you can leave this.

PR #1270, opened 2025-07-07
feat(engine): add cache store for nodes to NodeContext

rot1024:
# Overview

## What I've done

Add cache store for node to improve performance of node processing


pyshx:
Since we're already working on incremental Runs, we don't need this.

PR #1280, opened 2025-07-14
fix(api): stop log watching when job is done[FLOW-BE-170]

kasugamirai:
# Overview

This PR enhances the LogInteractor component to include job repository integration and implements job status monitoring. The primary goal is to optimize resource usage by automatically stopping log monitoring for jobs that have reached terminal states (completed, failed, or cancelled), preventing unnecessary polling and resource consumption.


pyshx (approved):
Makes sense! Approved.

PR #1284, opened 2025-07-21
refactor(ui): Tabular and Maplibre Optimizations [FLOW-FE-215]

billcookie:
# Overview
This PR aims to improve the performance of both the Tabular and Maplibre viewers for when a user has run a debug workflow.
## What I've done

Tabular View:

- Added small fix for tanstack virtualization for smoother scrolling

Maplibre
- Changed from markers to points/layers to allow more points to be shown at the same time.

### DIFF
--- a/ui/src/components/visualizations/MapLibre/index.tsx
+++ b/ui/src/components/visualizations/MapLibre/index.tsx
@@ -77,17 +77,13 @@ const MapLibre: React.FC<Props> = ({ className, fileContent, fileType }) => {
         mapStyle="https://basemaps.cartocdn.com/gl/positron-gl-style/style.json"
         style={{ width: "100%", height: "100%" }}
         maplibreLogo={true}
-        interactiveLayerIds={["polygon-layer", "line-layer"]}
+        interactiveLayerIds={["point-layer", "line-layer", "polygon-layer"]}
         onClick={(e) => {
           setSelectedFeature(e.features?.[0]);
         }}
         onLoad={handleMapLoad}>
         {fileType === "geojson" && (
-          <GeoJsonDataSource
-            fileType={fileType}
-            fileContent={fileContent}
-            onSelectedFeature={setSelectedFeature}
-          />
+          <GeoJsonDataSource fileType={fileType} fileContent={fileContent} />
         )}
       </Map>
       {selectedFeature && (
--- a/ui/src/features/Editor/components/OverlayUI/components/DebugPanel/index.tsx
+++ b/ui/src/features/Editor/components/OverlayUI/components/DebugPanel/index.tsx
@@ -28,7 +28,7 @@ import { useT } from "@flow/lib/i18n";
 
 import DebugLogs from "./DebugLogs";
 import DebugPreview from "./DebugPreview";
-import { DataTable } from "./DebugPreview/components";
+import DataTable from "./DebugPreview/components/DataTable";
 import useHooks from "./hooks";
 
 const DebugPanel: React.FC = () => {
--- a/ui/src/components/DataTable/index.tsx
+++ b/ui/src/components/DataTable/index.tsx
@@ -236,7 +236,8 @@ function DataTable<TData, TValue>({
         <div className="overflow-auto rounded-md border">
           <div
             ref={parentRef}
-            className="h-full overflow-auto rounded-md border">
+            className="h-full overflow-auto rounded-md border"
+            style={{ contain: "paint", willChange: "transform" }}>
             <Table>
               <TableHeader>
                 {table.getHeaderGroups().map((headerGroup) => (
--- a/ui/src/components/visualizations/MapLibre/sources/GeoJson/index.tsx
+++ b/ui/src/components/visualizations/MapLibre/sources/GeoJson/index.tsx
@@ -1,5 +1,5 @@
 import { useMemo } from "react";
-import { Source, Marker, Layer, LayerProps } from "react-map-gl/maplibre";
+import { Source, Layer, LayerProps } from "react-map-gl/maplibre";
 
 type Props = {
   fileType: "geojson";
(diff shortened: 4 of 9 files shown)


ui/src/components/visualizations/MapLibre/index.tsx
```diff
@@ -41,59 +75,119 @@ const MapLibre: React.FC<Props> = ({ className, fileContent, fileType }) => {
...
-          setSelectedFeature(e.features?.[0]);
+          if (e.features?.[0]) {
+            onSelectedFeature(e.features[0]);
+            setShowFeaturePanel(true);
+          }
+        }}
+        onDblClick={(e) => {
+          if (e.features?.[0]) {
+            onSelectedFeature(e.features[0]);
+            onShouldFlyToFeatureChange?.(true);
```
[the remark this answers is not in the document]
KaWaite: Actually, even if I click (not double) I get the panel, so your duplicating/mixing logic around this. Please double check where the feature panel is being toggled on/off and make sure nothing conflicts/contradicts

ui/src/components/visualizations/MapLibre/index.tsx
```diff
@@ -41,59 +75,119 @@ const MapLibre: React.FC<Props> = ({ className, fileContent, fileType }) => {
...
-    if (mapRef.current && dataBounds) {
-      mapRef.current.fitBounds(dataBounds, {
-        padding: 40,
-        duration: 0,
-        maxZoom: 16,
-      });
+  const convertedSelectedFeature = useMemo(() => {
+    if (!selectedFeature || !featureMap) return null;
+
+    if ("geometry" in selectedFeature && selectedFeature.geometry) {
```
KaWaite, line 81: All features should have geometry. In what case would this fail?

ui/src/components/visualizations/MapLibre/index.tsx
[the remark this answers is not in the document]
billcookie: I have intentionally added both as unsure what we want for the UX. Do we want it only on single, double or both?

ui/src/utils/geojson/parseGeoJson.ts
KaWaite, line 1: Just a note to put it in your head this file will probably need to become a function prepping data for maplibre, rather than something specific to geojson. After you read this you can resolve this.

ui/src/features/Editor/components/OverlayUI/components/DebugPanel/index.tsx
```diff
@@ -53,6 +53,30 @@ const DebugPanel: React.FC = () => {
   } = useHooks();
   const t = useT();
   const [tabValue, setTabValue] = useState("debug-logs");
+
+  const [selectedFeature, setSelectedFeature] = useState<any>(null);
+  const [enableClustering, setEnableClustering] = useState<boolean>(true);
+  const [shouldFlyToFeature, setShouldFlyToFeature] = useState<boolean>(false);
```
KaWaite, line 59: The way you are using this `shouldFlyToFeature` is a very precarious approach and should be avoided if possible.
Since maplibre accepts a Ref, you can define it high enough that you can use its methods inside of callbacks. Then, pass the ref down into Maplibre, and from the ref use fly to inside a callback that'll be run on the double click. I know this is a little tricky so if you don't quite understand I can try to explain on zoom and/or we can pair program it.

ui/src/components/visualizations/MapLibre/index.tsx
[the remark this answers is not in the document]
KaWaite: If a double click will mean the onClick isn't fired, what you have is fine. But if by double clicking you get onClick and onDblClick events to fire, then you can remove the duplicate code from the onDblClick and keep only what is not already happening on the onClick.

ui/src/components/visualizations/MapLibre/index.tsx
```diff
@@ -41,59 +75,122 @@ const MapLibre: React.FC<Props> = ({ className, fileContent, fileType }) => {
...
           if (instance) mapRef.current = instance.getMap();
         }}
         mapStyle="https://basemaps.cartocdn.com/gl/positron-gl-style/style.json"
         style={{ width: "100%", height: "100%" }}
         maplibreLogo={true}
-        interactiveLayerIds={["polygon-layer", "line-layer"]}
+        interactiveLayerIds={["point-layer", "line-layer", "polygon-layer"]}
         onClick={(e) => {
-          setSelectedFeature(e.features?.[0]);
+          console.log("Map clicked:", e.features);
```
KaWaite, line 163: Remove console log

ui/src/hooks/useDataColumnizer.ts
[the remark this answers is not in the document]
billcookie: For this I am a little confused with this. I thought we want originalId on both the tabular and 2D viewer? Is that now not the case? Or you mean you do not want the originalId to be showing on the tabular view or do not want it to be?

ui/src/hooks/useDataColumnizer.ts
```diff
@@ -38,6 +38,7 @@ export default ({
         });
 
         // Create columns for table
+
```
KaWaite, line 41: Remove added space

ui/src/features/Editor/components/OverlayUI/components/DebugPanel/DebugPreview/hooks.ts
```diff
@@ -0,0 +1,39 @@
...
+import { useCallback, useMemo } from "react";
+
+export default ({
+  mapRef,
+  selectedOutputData,
+}: {
+  mapRef: any;
+  selectedOutputData: any;
+}) => {
+  const dataBounds = useMemo(() => {
```
KaWaite, line 12: Since this only seems to be used in handleMapLoad, just move that code into there

ui/src/components/visualizations/MapLibre/index.tsx
```diff
@@ -77,24 +56,39 @@ const MapLibre: React.FC<Props> = ({ className, fileContent, fileType }) => {
...
+            onSelectedFeature(undefined);
+            setShowFeaturePanel(false);
+          }
+        }}
+        onDblClick={(e) => {
+          if (e.features?.[0]) {
+            onSelectedFeature(e.features[0]);
+            onFlyToSelectedFeature?.(e.features[0]);
+            setShowFeaturePanel(true);
+          }
```
KaWaite, line 73: Can you move both onClick and onDblClick code into callbacks? (`handleMapClick`, `handleMapDoubleClick`)

KaWaite (approved):

PR #1292, opened 2025-07-29
feat(ui): CMS Integration Mock Server/Data [FLOW-FE-226]

billcookie:
# Overview
The PR adds support for CMS Projects via GraphQL and Mock Data

## What I've done
- Added GraphQL Requests for CMS Projects, Models and Items
- Added Mock Server/Data integration for testing and UI implementation


ui/src/types/cms.ts
KaWaite, line 1: `cmsIntegration.ts` is maybe more informative.

ui/src/lib/gql/convert.ts
```diff
@@ -164,6 +175,52 @@ export const toAsset = (asset: AssetFragment): Asset => ({
     : "pending",
 });
 
+export const toCmsProject = (cmsProject: CmsProjectFragment): CmsProject => ({
```
KaWaite, line 178: CMS should be capitalized

ui/src/lib/gql/convert.ts
```diff
@@ -164,6 +175,52 @@ export const toAsset = (asset: AssetFragment): Asset => ({
...
+  id: cmsProject.id,
+  workspaceId: cmsProject.workspaceId,
+  name: cmsProject.name,
+  alias: cmsProject.alias,
+  description: cmsProject.description ?? null,
+  license: cmsProject.license ?? null,
+  readme: cmsProject.readme ?? null,
```
KaWaite, line 185: With your type update to match the schema, you shouldn't need these fallback nulls anymore, so please update all converting functions appropriately to match types 🙏🏻

ui/src/mocks/data/cms.ts
KaWaite, line 1: Sorry I didn't notice in the mock setup PR, but is there a reason we are creating separate types and not just using our own types? Aren't they one-to-one? 
If there is no specific reason, please use the CMS types in types/cms.ts and in another PR if you can update all the other domains to use the actual types 🙏🏻

billcookie: I have been looking through why this is the case and it was done on the original setup. I think because the data wont be exactly one for one especially around enums. For example, we need to replicate raw data for the cmsProject which means instead of 'public' which is what we want the CmsProject type to be we should set it as "PUBLIC" for the convert.ts file to work properly

ui/src/lib/gql/convert.ts
billcookie: Sorry should this apply across anything that has cms across the project? e.g. cmsProject should become CMSProject and all types the same? I had started with all capitalised but then reverted back

ui/src/mocks/data/cms.ts
KaWaite: Ah okay, I was looking at it reversed. So I think since we are generating our graphql types/etc from the api schema, I tihnk we can (and should) use those types. So same as what we use in the useQuery files.

ui/src/features/AssetsDialog/index.tsx
```diff
@@ -75,7 +76,24 @@ const AssetsDialog: React.FC<Props> = ({
...
+  const { page } = useGetCmsProjects("ws-001", false);
+  const { page: models } = useGetCmsModels("proj-001");
+  const { cmsProject } = useGetCmsProject("proj-001");
+  const { page: items } = useGetCmsItems("proj-001", "model-001");
+  const { page: exportUrl } = useGetCmsModelExportUrl("proj-001", "model-001");
+  console.log("CMS Projects:", page?.cmsProjects);
+  console.log("CMS MODELs:", models?.cmsModels);
+  console.log("CMS Project:", cmsProject);
+  console.log("CMS ITEMS:", items?.cmsItems);
+  console.log("CMS EXPORT URL:", exportUrl?.cmsModelExportUrl);
```
KaWaite, line 96: Remove console logs :)

ui/src/features/AssetsDialog/index.tsx
KaWaite, line 1: I don't think this file should be in this PR

PR #1334, opened 2025-08-18
fix(ui): Saving Project, Deleting Asset and Debounced Search Fix [FLOW-FE-246]

billcookie:
# Overview
PR adds a saving state that prevents users to be able to spam the save, delete an asset more than once and fixes debounced search
## What I've done
- Added a shared state for manual saving a project by lifting logic upwards
- Asset actions are now disabled when an asset is being deleted
- Debounced search triggers faster on when a user starts to type (Assets)

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -5741,7 +5741,7 @@ checksum = "7edddbd0b52d732b21ad9a5fab5c704c14cd949e5e9a1ec5929a24fded1b904c"
 
 [[package]]
 name = "plateau-gis-quality-checker"
-version = "0.0.90"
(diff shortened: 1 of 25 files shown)


ui/src/features/AssetsDialog/hooks.ts
```diff
@@ -197,6 +206,8 @@ export default ({
   return {
     assets,
     isFetching,
+    isDebouncing,
```
KaWaite, line 209: isDebouncingSearch

KaWaite (approved):

KaWaite (approved):

PR #1340, opened 2025-08-18
feat(engine): add shapefile support with reader implementation

pyshx:
# Overview
Add support for importing ESRI Shapefile format (.shp) to the Re:Earth Flow engine

## What I've done
 - Implemented shapefile reader that reads from ZIP archives containing .shp, .dbf, and optionally .shx files
 - Added support for all common shapefile geometry types (Point, Polyline, Polygon with 2D/3D variants)
- Integrated with existing FileReader infrastructure following established patterns
- Converts shapefile geometries to internal CityGmlGeometry format
- Preserves all DBF attributes as feature properties
- All processing done in-memory without temporary files

## What I haven't done
- Multipoint and Multipatch geometry types are not yet supported (returns error)
- Character encoding support is limited to UTF-8 (encoding parameter exists but not implemented)
- Direct shapefile bytes (non-ZIP) are not supported - requires ZIP archive

## How I tested
- Unit tests for ZIP detection, geometry conversion, and attribute conversion
- Integration tested with real-world data:
  - US Census Bureau states shapefile (56 polygon features, 2.7MB)
  - Created test shapefiles with pyshp library
 - Verified geometry output using GeoJsonWriter
 - Confirmed both attributes and geometry are correctly preserved
- All tests passing: cargo test -p reearth-flow-action-source shapefile
- No clippy warnings: cargo clippy -p reearth-flow-action-source

1. Geometry conversion approach - Using CityGmlGeometry as the target format (consistent with other readers in the codebase)
2. Error handling - Whether the error messages are descriptive enough
3. Memory efficiency - Reading entire ZIP into memory buffers (alternative would be streaming but more complex)

## Memo
- CityGmlGeometry is used as the primary geometry format throughout the engine, not just for CityGML files
- The .shx index file is read but not currently utilized (could be used for random access in future)

### DIFF
--- a/engine/docs/mdbook/src/action.md
+++ b/engine/docs/mdbook/src/action.md
@@ -2520,51 +2520,6 @@ Reads features from a file
           ]
         }
       }
-    },
-    {
-      "title": "Shapefile",
-      "type": "object",
-      "required": [
-        "format"
-      ],
-      "properties": {
-        "dataset": {
-          "title": "File Path",
-          "description": "Expression that returns the path to the input file (e.g., \"data.csv\" or variable reference)",
-          "anyOf": [
-            {
-              "$ref": "#/definitions/Expr"
-            },
-            {
-              "type": "null"
-            }
-          ]
-        },
-        "encoding": {
-          "type": [
-            "string",
-            "null"
-          ]
-        },
-        "format": {
-          "type": "string",
-          "enum": [
-            "shapefile"
-          ]
-        },
-        "inline": {
-          "title": "Inline Content",
-          "description": "Expression that returns the file content as text instead of reading from a file path",
-          "anyOf": [
-            {
-              "$ref": "#/definitions/Expr"
-            },
-            {
-              "type": "null"
-            }
-          ]
-        }
-      }
     }
   ],
   "definitions": {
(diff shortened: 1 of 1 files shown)


engine/runtime/action-source/src/file/reader/runner.rs
```diff
@@ -63,6 +63,13 @@ pub enum FileReader {
         #[serde(flatten)]
         common_property: FileReaderCommonParam,
     },
+    /// # Shapefile
+    Shapefile {
+        #[serde(flatten)]
+        common_property: FileReaderCommonParam,
+        #[serde(flatten)]
+        property: shapefile::ShapefileReaderParam,
+    },
```
n4to4, line 72: Could you implement this as an independent `ShapefileReader` instead of as part of `FileReader`?

We are planning to deprecate the implementation in `FileReader`, and the plan is to make all of them independent `Reader`s in the same way.

For reference, please take a look at [this PR](https://github.com/reearth/reearth-flow/pull/1192/files).

n4to4 (changes requested):

n4to4 (approved):
LGTM

PR #1345, opened 2025-08-19
feat(engine): add CityGML geometry support, fix decimal precision for Clipper action

rot1024:
## Summary
- Add full CityGML geometry support to the Clipper action
- Support composite surfaces (MultiSurface) with recursive processing
- Handle line strings (Curve type geometries) appropriately

## Changes
- Enable CityGML geometries as both clipper and candidate inputs
- Extract 3D polygons from CityGML for clipping operations
- Add recursive processing for composite surfaces (nested GML geometries)
- Handle line strings for Curve type geometries (preserved as-is since they cannot be clipped)
- Preserve CityGML metadata (materials, textures, UV coordinates) after clipping
- Add comprehensive tests for all CityGML geometry types (Solid, Surface, Curve, composite surfaces)
- Add documentation comments about 3D clipping limitations (2.5D surface clipping, not volumetric)

## Test Coverage
Added 18 tests covering:
- Basic 2D/3D polygon clipping
- CityGML as candidate geometry
- CityGML as clipper geometry  
- Composite surfaces (nested MultiSurface)
- Curve type with line strings
- Solid type with nested surfaces
- Geometry collections

All tests passing ✅

## Notes
- The 3D clipping is surface-level (2.5D), not true volumetric operations, as documented in code comments
- Line strings in Curve geometries are preserved as-is since they cannot be clipped like polygons

### DIFF
--- a/engine/runtime/action-processor/src/geometry/clipper.rs
+++ b/engine/runtime/action-processor/src/geometry/clipper.rs
@@ -362,6 +362,26 @@ fn forward_polygon3d(
     }
 }
 
+/// The scaling factor used to preserve decimal precision in Clipper operations.
+/// Clipper library internally uses integer arithmetic, so we scale up coordinates
+/// to preserve decimal places, then scale back down after processing.
+/// 
+/// Current setting: 10^8 preserves 8 decimal places of precision.
+/// 
+/// Precision vs Range tradeoff:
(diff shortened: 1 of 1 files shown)


n4to4 (approved):
LGTM

PR #1351, opened 2025-08-21
feat(engine): gltf compression

shunski:
# Overview
Draco's glTF compression added to the cesium3DTileWriter. The compression option can be configured as a parameter of the cesium3DTilesWriter and is on by default.


n4to4 (approved):
LGTM 👍

PR #1357, opened 2025-08-22
feat(server): replace user management to accounts for SearchUser [FLOW-BE-191]

akiyatomohiro:
# Overview


server/api/internal/usecase/interfaces/user.go
```diff
@@ -9,4 +9,5 @@ import (
 
 type User interface {
 	FindByIDs(context.Context, id.UserIDList) (user.List, error)
+	UserByNameOrEmail(context.Context, string) (*user.User, error)
```
akiyatomohiro, line 12: As for the name, I was torn between userByNameOrEmail and searchUser. Since we use the accounts server’s userByNameOrEmail schema, I went with the former.

PR #1360, opened 2025-08-25
fix(ui): added react-query for debug run data caching [FLOW-FE-247]

billcookie:
# Overview
Debug run output urls/intermediate data was not being cached meaning data was being fetched each time.
## What I've done
This PR uses react-query to allow for caching for those results to prevent unnecessary fetching.

### DIFF
--- a/ui/src/hooks/useDebugRunUrlQuery.ts
+++ b/ui/src/hooks/useDebugRunUrlQuery.ts
@@ -8,8 +8,8 @@ export const useDebugRunUrlQuery = (dataUrl: string) => {
     queryFn: () => {
       return fetchAndReadData(dataUrl);
     },
-    staleTime: 5 * 60 * 1000,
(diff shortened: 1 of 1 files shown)


ui/src/hooks/useDebugRunUrlQuery.ts
```diff
@@ -0,0 +1,16 @@
...
+import { fetchAndReadData } from "@flow/utils/fetchAndReadGeoData";
+
+export const useDebugRunUrlQuery = (dataUrl: string) => {
+  return useQuery({
+    queryKey: ["dataUrl", dataUrl],
+    queryFn: () => {
+      return fetchAndReadData(dataUrl);
+    },
+    staleTime: 5 * 60 * 1000,
+    gcTime: 10 * 60 * 1000,
```
KaWaite, line 12: Any reason these are set as they are and not just set to infinity?

PR #1377, opened 2025-08-28
feat(engine): enhance CzmlReader to support additional geometries including ellipses and rectangles

pyshx:
# Overview
Enhance CzmlReader to support additional geometries

## What I've done
CZMLReader enhancements
  - Rectangle support: Converts `wsenDegrees/wsenRadians` bounds to polygon
  - Ellipse support: Approximates as 33-point polygon using semi-major/minor axes
  - Polygon holes support: Properly extracts and preserves interior rings
  - Corridor support: Extracts as LineString (proper buffering would need complex geometry ops)
  - Wall support: Extracts as 3D LineString with elevation
  - Fixed geometry extraction priority: Specific geometries checked before falling back to position field

1. Ellipse approximation accuracy: Using rough meter-to-degree conversion (meters_per_degree_lon = 111_000 * cos(lat)). Should use proper geodesic calculations for
  production.
2. Position field handling in CZML Reader: Changed to check specific geometries first before position. This prevents ellipses/boxes from being reduced to points but might
   affect some edge cases.
3. Corridor as LineString: Currently just extracts positions without buffering. Proper corridor polygon would require complex geometry buffer operations.

### DIFF
--- a/engine/runtime/action-source/src/file/czml.rs
+++ b/engine/runtime/action-source/src/file/czml.rs
@@ -240,7 +240,7 @@ fn packet_to_feature(
 fn extract_geometry(
     packet: &Value,
     force_2d: bool,
-) -> Result<Option<Geometry>, crate::errors::SourceError> {    
+) -> Result<Option<Geometry>, crate::errors::SourceError> {
     if let Some(polygon) = packet.get("polygon") {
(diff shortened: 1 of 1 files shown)


n4to4 (approved):
👍

PR #1385, opened 2025-08-31
feat(websocket): add dynamic sleep duration based on awareness update count and increase batch size to 500 [FLOW-BE-205]

kasugamirai:
# Overview
This PR optimizes Redis stream consumption performance by implementing dynamic sleep duration and increasing batch processing capacity.
Prevent Redis stream from being consumed too aggressively
 Reduce unnecessary polling when there are fewer updates


pyshx (approved):
LGTM

PR #1393, opened 2025-09-03
fix(ui): Keyboard shortcuts [FLOW-FE-241]

billcookie:
# Overview

Replaced keyboard shortcuts with react-hot-keys package instead for accurate use and to prevent traditional browser actions where possible.

For example, if you copy a node in the canvas and then copy text in the TabularView. These will be two different states of copy and will now not cause issue with one another.

## What I've done

- Integration of react-hotkeys-hook package for keyboard shortcut handling
- Removed some unused shortcuts and translations that are currently no longer relevant
- Enabled keyboard shortcuts for zoom operations, fullscreen toggle, and editor actions
- Consistent hotkey patterns across different components (main editor, shared views, user menu)

## Memo
Currently have kept the useShortcut hook as we may want to make some form of custom integration in the future.


ui/src/lib/i18n/locales/en.json
KaWaite, line 1: Why did we loose this language?

ui/src/features/KeyboardShortcutDialog/hooks.ts
```diff
@@ -51,14 +51,14 @@ export default () => {
...
-        description: t("Toggle the Actions List Panel"),
-      },
+      // {
+      //   keyBinding: EditorKeyBindings["leftPanelCanvasNavigator"],
+      //   description: t("Toggle the Canvas Navigator Panel"),
+      // },
+      // {
+      //   keyBinding: EditorKeyBindings["leftPanelActionsList"],
+      //   description: t("Toggle the Actions List Panel"),
+      // },
```
KaWaite, line 61: These can just be deleted since they are from old UI

KaWaite (approved):

KaWaite (approved):

PR #1405, opened 2025-09-08
feat(ui): add spotlight user functionality in collaboration components [FLOW-FE-245]

billcookie:
# Overview
This PRs adds the ability to follow another users cursor. When hovered a user in the collaborators list, an icon will appear to allow a user to follow another user. 
## Screenshot
Screenshot 2025-09-08 at 16 10 19
Screenshot 2025-09-08 at 16 10 35

### DIFF
--- a/ui/src/features/Editor/hooks.ts
+++ b/ui/src/features/Editor/hooks.ts
@@ -62,9 +62,7 @@ export default ({
     zoom,
   });
 
-  useEffect(() => {
-    latestViewportRef.current = { x, y, zoom };
-  }, [x, y, zoom]);
+  latestViewportRef.current = { x, y, zoom };
 
   const [spotlightUserClientId, setSpotlightUserClientId] = useState<
     number | null


ui/src/features/Editor/hooks.ts
```diff
@@ -379,10 +382,31 @@ export default ({
     (event: MouseEvent) => {
       if (yAwareness) {
         throttledMouseMove(event, yAwareness, screenToFlowPosition);
+        yAwareness.setLocalStateField("zoomLevel", getZoom());
```
KaWaite, line 385: Does throttling the zoom cause any issues? Best would be moving this into the throttledMouseMove, and also add a conditional check to see if the zoom level has changed (if not, don't update yAwareness). 
Because currently we update zoom level EVERY time, which means every pixel I think since its not throttled. Also, even if zoom doesn't change, we try to set it. Which I think yjs might be smart enough to not do anything in that case, but better to avoid if possible.

ui/src/features/Editor/hooks.ts
```diff
@@ -53,7 +53,22 @@ export default ({
...
+
+  const latestViewportRef = useRef<{ x: number; y: number; zoom: number }>({
+    x,
+    y,
+    zoom,
+  });
+
+  useEffect(() => {
+    latestViewportRef.current = { x, y, zoom };
+  }, [x, y, zoom]);
```
KaWaite, line 67: Does this need to be in a useEffect, or can it just be 
```suggestion
  latestViewportRef.current = { x, y, zoom };
```

PR #1435, opened 2025-09-15
feat(engine): enhance OBJ reader to support material properties

pyshx:
# Overview


engine/runtime/action-source/src/file/obj.rs
```diff
@@ -922,4 +1060,59 @@ f 4 5 6
...
+            diffuse: Some([1.0, f32::NAN, 0.5]),
+            shininess: Some(f32::INFINITY),
+            ..Default::default()
+        };
+        materials.insert("test_mat".to_string(), mat);
+
+        let used_materials = vec!["test_mat".to_string()];
+        let (_, material_properties) = extract_material_properties(&materials, &used_materials);
+
+        if let AttributeValue::Map(props) = material_properties {
```
n4to4, line 1104: I think replacing the `if let` patterns with `let...else` would be better, because the current approach can silently pass if the expected structure isn't found.

```rust
let AttributeValue::Map(props) = material_properties else {
    panic!("Expected material_properties to be a Map");
};

[...]
```

This way the test will explicitly fail with a clear message instead of potentially passing without testing anything. What do you think?

PR #1436, opened 2025-09-15
feat(ui): added subworkflow create via node/edge selection [FLOW-FE-121]

billcookie:
# Overview
This PR implements the ability to create a subworkflow from selected nodes and edges in the UI editor. Users can now select nodes and edges, then press the 's' key to extract them into a new subworkflow.
## What I've done

- Uncommented and activated the handleYWorkflowAddFromSelection function
- Updated the keyboard shortcut handling to trigger subworkflow creation with the shift + cmd + 's' key
- Modified the data structure handling to use maps instead of arrays for nodes and edges

### DIFF
--- a/ui/src/lib/yjs/useYWorkflow.ts
+++ b/ui/src/lib/yjs/useYWorkflow.ts
@@ -157,17 +157,6 @@ export default ({
     ],
   );
 
-  // Helper function to create a meaningful handle name
-  const getHandleName = (
-    node: Node | undefined,
-    handle: string | undefined | null,
-  ): string => {
-    if (!handle) {
-      return node?.data.officialName || "default";
-    }
-    return handle;
-  };
-
   const handleYWorkflowAddFromSelection = useCallback(
     async (nodes: Node[], edges: Edge[]) => {
       try {
@@ -182,6 +171,7 @@ export default ({
           if (containsReadersOrWriters) {
             return;
           }
+
           const nodesByParentId = new Map<string, Node[]>();
           nodes.forEach((node) => {
             if (node.parentId) {
@@ -668,4 +658,15 @@ export default ({
     handleYWorkflowRename,
     handleYWorkflowAddFromSelection,
   };
+};
+
+// Helper function to create a meaningful handle name
+const getHandleName = (
+  node: Node | undefined,
+  handle: string | undefined | null,
+): string => {
+  if (!handle) {
+    return node?.data.officialName || "default";
+  }
+  return handle;
 };


ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -86,11 +87,9 @@ export default ({
...
       };
 
-      const workflowNodes = [
-        newInputNode,
-        ...(initialNodes ?? []),
-        newOutputNode,
-      ];
+      const workflowNodes = needsDefaultRouters
+        ? [newInputNode, ...(initialNodes ?? []), newOutputNode]
+        : [...(initialNodes ?? [newInputNode, newOutputNode])];
```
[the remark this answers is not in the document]
billcookie: Logic implemented is correct. There are three possible conditions we are aiming for. 

1. Subworkflow creation with nodes that do not have boundary edges, but may have internal edges or no internal edges but we need to add default nodes.
2. Subworkflow creation that has external edges and we just pass the initialNodes as we create the input and output nodes outside of this creation as it is easier to assign them on creation rather than linking nodes afterwards.
3. When we create a subworkflow without any initialNodes e.g. dragging from the action bar.

ui/src/features/Editor/components/OverlayUI/components/NodePickerDialog/index.tsx
```diff
@@ -64,6 +65,7 @@ const NodePickerDialog: React.FC<Props> = ({
           />
           <Select
             value={currentActionByType}
+            disabled={currentActionByType === "transformer" && !isMainWorkflow}
```
KaWaite: I think the second suggestion is best and more logically accurate. We don't want things disabled if the currentActionByType is transformer and then is a subworkflow. We want it if the action is writer or reader and in a subworkflow

ui/src/lib/yjs/useYWorkflow.ts
```diff
@@ -158,121 +157,444 @@ export default ({
...
+          const selectedNodes = nodes.filter((n) => n.selected);
+          if (selectedNodes.length === 0) return;
+
+          const containsReadersOrWriters = selectedNodes?.some(
+            (n) => n.type === "reader" || n.type === "writer",
+          );
+          if (containsReadersOrWriters) {
+            return;
+          }
+          const nodesByParentId = new Map<string, Node[]>();
```
KaWaite:
```suggestion
          }

          const nodesByParentId = new Map<string, Node[]>();
```

PR #1438, opened 2025-09-16
fix(engine): csv tests for unit tests and some output file renamings

shunski:
# Overview

## What I've done
- Unit tests can now handle csv files.
- Unit tests do not take comparison methods anymore. They are automatically computed from the file formats.
- Renamed the output filenames for the plateau workflow and the corresponding test data.


n4to4 (approved):
👍

PR #1475, opened 2025-09-25
feat(engine): plateau4 qc 02 solid boundary validator

shunski:
# Overview

## What I've done
- Implemented SolidBoundaryValidator.
    - checks for orientability, connectibity, 2-manifold condition, face orientation.
    - It assumes all the surface validation is already done. It sends the feature to the rejected port whenever it detects the surface issues that it is not in charge of.
- Modified the unit test system so that it can handle multiple expected output files instead of just one.
- Added some helper objects and algorithms in engine/runtime/geometry in order to implement the SolidBoundaryValidator.


n4to4 (approved):
LGTM 👍

PR #1480, opened 2025-09-29
feat(ui): disable node prop added [FLOW-FE-270]

billcookie:
# Overview
This PR adds a disable/enable functionality for workflow nodes, allowing users to exclude specific nodes from workflow execution without deleting them. User can right click on a node to disable it. 
## What I've done

- Adds isDisabled property to node data structure
- Implements visual indication for disabled nodes with opacity and special colors
- Updates workflow conversion logic to filter out disabled nodes during execution

## Screenshot
Screenshot 2025-09-29 at 15 18 23
Screenshot 2025-09-29 at 15 18 39

### DIFF
--- a/ui/src/components/SchemaForm/Templates/ArrayFieldItemTemplate.tsx
+++ b/ui/src/components/SchemaForm/Templates/ArrayFieldItemTemplate.tsx
@@ -40,7 +40,7 @@ const ArrayFieldItemTemplate = <
         <div className="flex-1">{children}</div>
         {hasToolbar && (
           <div
-            className={`${schema.required ? "absolute top-0 right-0" : ""} flex items-center gap-1`}>
+            className={`${schema.required ? "absolute top-3.5 right-0" : ""} flex items-center gap-1`}>
             {(hasMoveUp || hasMoveDown) && (
               <MoveUpButton
                 disabled={readonly || disabled || !hasMoveUp}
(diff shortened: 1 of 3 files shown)


ui/src/lib/reactFlow/nodeTypes/GeneralNode/index.tsx
```diff
@@ -39,10 +39,10 @@ const GeneralNode: React.FC<GeneralNodeProps> = ({
 
   return (
     <div
-      className={`max-w-[200px] min-w-[150px] rounded-md border bg-secondary ${selected ? selectedColor : borderColor}`}>
+      className={`max-w-[200px] min-w-[150px] rounded-md border bg-secondary ${selected ? selectedColor : borderColor} ${data.isDisabled ? "opacity-70" : ""}`}>
       <div className="relative m-1 flex h-[25px] items-center gap-1 rounded-sm">
         <div
-          className={`flex justify-center self-center rounded-sm border p-1 align-middle ${selected ? selectedColor : borderColor} ${selected ? selectedBackgroundColor : className}`}>
+          className={`flex justify-center self-center rounded-sm border p-1 align-middle ${selected ? selectedColor : borderColor} ${selected ? selectedBackgroundColor : backgroundColor}`}>
```
KaWaite, line 45: If you are adding the backgroundColor like this, you can update the individual nodes that are using this GeneralNode to not pass it anymore

KaWaite: But also double check whether there are any other important classes now being missed due to you replacing className use

ui/src/features/Editor/hooks.ts
```diff
@@ -355,6 +356,29 @@ export default ({
     handleWorkflowClose,
   });
 
+  const handleNodeDisable = useCallback(
```
KaWaite, line 359: Pretty sure we want this with the rest of the yjs code or else we start mixing domains again

PR #1491, opened 2025-09-30
fix(ui): subworkflow input and output router fix [FLOW-FE-280]

billcookie:
# Overview
This PR fixes an issue where pseudo input and output ports are deleted when all routers are removed and then a new one is added. The solution involves adding logic to initialize pseudo ports when creating input or output routers in subworkflows.
## What I've done

- Added a new function to handle pseudo port creation when adding router nodes
- Modified the node creation callback to call the new pseudo port handling function
- Updated dependencies in the useCallback hook to include required workflows data

### DIFF
--- a/ui/src/lib/yjs/useYNode.ts
+++ b/ui/src/lib/yjs/useYNode.ts
@@ -41,8 +41,8 @@ export default ({
           }
 
           // For routers without routingPort, generate unique port name
-          const isRouterInput = newNode.data.outputs?.length;
-          const isRouterOutput = newNode.data.inputs?.length;
+          const isRouterInput = newNode.data.officialName === "InputRouter";
+          const isRouterOutput = newNode.data.officialName === "OutputRouter";
 
           if (
             (isRouterInput || isRouterOutput) &&


ui/src/lib/yjs/useYNode.ts
```diff
@@ -38,11 +39,101 @@ export default ({
               return [...snids, newNode.id];
             });
           }
+
+          // For routers without routingPort, generate unique port name
+          const isRouterInput = newNode.data.outputs?.length;
+          const isRouterOutput = newNode.data.inputs?.length;
```
KaWaite:
```suggestion
          const hasRouterInput = !!newNode.data.outputs?.length;
          const hasRouterOutput = !!newNode.data.inputs?.length;
```

ui/src/lib/yjs/useYNode.ts
```diff
@@ -38,11 +39,101 @@ export default ({
...
+
+          if (
+            (isRouterInput || isRouterOutput) &&
+            !newNode.data.params?.routingPort
+          ) {
```
KaWaite: I am not sure this is sufficient for knowing whether we are dealing with an OutputRouter or InputRouter, is it? Won't most nodes being added meet these conditions with at least outputs if not inputs? 🤔 
If so, that means this check is pointless

KaWaite (approved):

PR #1500, opened 2025-10-03
refactor(engine): extract JsonReader and CityGmlReader from FileReader as separate actions

asrcpq:
# Overview

This PR completes #1192 for splitting JSON and CityGML reader from the generic FileReader.


engine/runtime/action-source/src/file/citygml.rs
```diff
@@ -0,0 +1,121 @@
...
+        ctx: NodeContext,
+        sender: Sender<(Port, IngestionMessage)>,
+    ) -> Result<(), BoxedError> {
+        let storage_resolver = Arc::clone(&ctx.storage_resolver);
+        let input_path = self
+            .params
+            .common_property
+            .dataset
+            .as_ref()
+            .and_then(|_| None); // Placeholder, see below
```
n4to4, line 110: Could you just use `runner::get_input_path`, and make it `pub(crate)` if needed?

n4to4 (approved):
LGTM 👍

PR #1502, opened 2025-10-03
feat(engine): add support for converting polygons to geometry in shapefileReader and update dependencies

pyshx:
# Overview


engine/runtime/action-source/src/file/shapefile.rs
```diff
@@ -283,6 +284,222 @@ fn read_shapefile_from_zip(
...
+
+    let mut polygon_data = Vec::new();
+    let mut current_exterior: Option<Vec<Coordinate<f64, NoValue>>> = None;
+    let mut current_holes: Vec<Vec<Coordinate<f64, NoValue>>> = Vec::new();
+
+    for ring in rings {
+        match ring {
+            PolygonRing::Outer(points) => {
+                if let Some(exterior) = current_exterior.take() {
+                    polygon_data.push((exterior, current_holes.clone()));
```
n4to4, line 313: Same here:
https://github.com/reearth/reearth-flow/pull/1503/files#r2403175798

PR #1534, opened 2025-10-10
fix(engine): fix cross-device file moving

asrcpq:
# Overview

This PR generalize rename to copy files when they are on different devices. This fixes workflow-tests failing on Linux which uses tmpfs as the default cache dir and fail with the `rename`.

Semantically, unlike `rename`, it is the standard to use copy fallback for `move_file` implementation in high-level filesystem libraries such as `fs_extra`'s `move_file`.

## Note

- The reason that CI passed without error is because CI uses Ubuntu 22.04 which disables systemd's `tmp.mount` so tmp is using root filesystem, not tmpfs. This will trigger error since Ubuntu 24.10. See https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/.
- Checked all other `fs::rename` call sites in reearth-flow and they are guaranteed to be same-device moving, not need to switch to `move_file`.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -6631,7 +6631,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-log"
-version = "0.0.400"
+version = "0.0.401"
 dependencies = [
  "futures",
  "once_cell",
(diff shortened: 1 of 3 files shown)


n4to4 (approved):
👍

PR #1969, opened 2026-03-09
fix(ui): Add triangular meshes and line support [FLOW-DEV-18]

billcookie:
# Overview
This PR adds Cesium rendering support for a new FlowGeometry3D geometry type that includes triangular meshes and line segments. It creates a new utility for converting FlowGeometry3D features into Cesium Primitives, a new React component to manage their lifecycle, and integrates triangular mesh support into the existing CityGML rendering pipeline.

## What I've done
- New `flowGeometry3DToPrimitives.ts` utility and `FlowGeometry3DData.tsx` component to handle `FlowGeometry3D` features with triangular meshes and line segments as batched Cesium Primitives.
- Extended `cityGmlGeometryToPrimitives.ts` and `CityGmlData.tsx` to support `triangularMesh` geometry within CityGML features (for both known 3D types and unknown types), including a new `meshPrimitive` draw call.
- Fixed `dataSources` iteration in `hooks.ts` to use indexed access (`.get(i)`) instead of `for...of`, which is required by CesiumJS's `DataSourceCollection` API.


ui/src/features/Editor/components/OverlayUI/components/DebugPanel/hooks.ts
```diff
@@ -281,7 +281,9 @@ export default () => {
               cesiumViewer.zoomTo(matchingEntities);
             } else {
               // Search in data sources as fallback
-              for (const dataSource of cesiumViewer.dataSources) {
+
+              for (let i = 0; i < cesiumViewer.dataSources.length; i++) {
+                const dataSource = cesiumViewer.dataSources.get(i);
```
KaWaite: Any reason for this change? 🤔

PR #2034, opened 2026-04-03
refactor(api, ui): move snapshot saving on debug run from ui to server within RunProject[FLOW-DEV-61]

KaWaite:
# Overview
Currently the frontend calls saveSnapshot before runProject as two separate mutations. This moves the snapshot flush server-side — RunProject now calls FlushToGCS directly before submitting the job, ensuring a snapshot is always saved atomically with the run. The frontend saveSnapshot call before runProject is removed accordingly.


billcookie (approved):

server/api/gql/job.graphql
```diff
@@ -1,7 +1,7 @@
 type Job implements Node {
   completedAt: DateTime
   deployment: Deployment
-  deploymentId: ID!
```
pyshx: Let's keep this optional as before.

pyshx (approved):

PR #2043, opened 2026-04-10
fix(engine): action schemas not having working internationalization support [FLOW-DEV-72]

KaWaite:
# Overview

Add flat property-path i18n support for action parameter schemas, and introduce a `scaffold-i18n` CLI command to keep translation files in sync automatically.

## What I've done

- Extended `I18nSchema` in `engine/cli/src/utils.rs` with `parameterI18n` and `definitionI18n` fields, replacing the unused JSON Merge Patch mechanism with a flat property-path walker
- Added `cargo make scaffold-i18n` — reconciles all `schema/i18n/actions/{lang}.json` files against the current action schemas (adds missing keys, removes stale keys, preserves existing translations)
- Ran `scaffold-i18n` to populate skeleton entries for all existing actions across all 5 language files
- Updated `engine/AGENTS.md` with the new workflow and i18n format documentation
- Add full Japanese support

## What I haven't done

- Filled in actual translations — the scaffolded entries are empty and ready for translators
- Added schemars `title`/`description` annotations to actions that are currently missing them (would improve English fallback strings)
- Non-Japanese translations are incomplete

## How I tested

Manually verified the full pipeline end-to-end:
- Added Japanese translations for `AreaCalculator` (parameter properties) and `AttributeManager` (definition properties)
- Ran `cargo make doc-action` and confirmed the translations appeared correctly in `actions_ja.json` while `actions.json` (base) remained unchanged
- Confirmed existing translations in all lang files were preserved after `scaffold-i18n`

## Screenshot

N/A

- The `reconcile_action` logic in `scaffold_i18n.rs` — specifically whether silently skipping stale keys (rather than warning) is the right behavior
- Whether `definitionI18n` covering only one level of `schema["definitions"]` is sufficient for current action schemas, or if any actions have definitions that reference other definitions

## Memo

The generated `schema/actions_*.json` files are still committed to git and fetched by the API server at startup — no changes to that pipeline. The server requires no changes since it forwards the `parameter` field as an opaque JSON object.


engine/AGENTS.md
```diff
@@ -56,11 +58,41 @@ Each action defines input/output ports, JSON schema for validation, and paramete
...
-6. Run `cargo make doc-action` to update schemas
+5. Run `cargo make schema-base` — regenerates `actions.json` and syncs i18n skeleton entries for the new action across all language files
+6. Fill in translated strings in `schema/i18n/actions/{lang}.json` for each language
+7. Run `cargo make schema-translated` — generates all `actions_{lang}.json` files and docs
+
+**Always run both commands in order. Never run `schema-translated` without first running `schema-base` when action code has changed.**
+
+### Action i18n
+
+Translated action schemas are generated from source files in `schema/i18n/actions/{lang}.json`. **Never edit the generated `schema/actions_*.json` files directly** — they are always overwritten by `cargo make doc-action`.
```
n4to4: The `doc-action` task no longer exists (replaced by `schema-translated`).

Similarly, `runtime/action-processor/README.md:223` still references `cargo make doc-action`.

engine/cli/src/scaffold_i18n.rs
```diff
@@ -0,0 +1,337 @@
...
+
+        let dir = Path::new(&self.dir);
+        if !dir.exists() {
+            return Err(crate::errors::Error::init(format!(
+                "i18n directory not found: {}",
+                dir.display()
+            )));
+        }
+
+        for entry in fs::read_dir(dir).map_err(|e| crate::errors::Error::init(format!("{e:?}")))? {
```
n4to4: `scaffold_i18n.rs:284-330` — All I/O errors are stringified via Debug format (`{e:?}`) before being passed to `Error::init()`. Debug output is verbose and not user-friendly. `Display` (`{e}`) would be cleaner, and adding the file path to the message would make errors actionable.

```rust
// Current
fs::read_to_string(&path).map_err(|e| crate::errors::Error::init(format!("{e:?}")))?;

// Suggested
fs::read_to_string(&path)
    .map_err(|e| crate::errors::Error::init(format!("failed to read {}: {e}", path.display())))?;
`

n4to4 (approved):
👍

PR #2049, opened 2026-04-14
perf(engine): spill features to disk during finish() to prevent shutdown deadlocks

pyshx:
## Summary
Add a spill-to-disk mechanism during accumulating processor `finish()` to prevent shutdown deadlocks on bounded channels.

## Problem
When accumulating processors (FeatureSorter, LineOnLineOverlayer, etc.) emit many features during `finish()`, bounded channels (2048 buffer) fill up and `send()` blocks indefinitely. The downstream can't drain because it's waiting for Terminate from another branch — causing 6-hour batch timeouts.

## Solution
Enable "spill mode" before `finish()`:
- `send()` switches to `try_send()` — non-blocking
- If channel full, features are written to a JSONL spill file on disk
- After `finish()` returns, spill files are flushed as `FileBackedOp` through existing infrastructure
- Downstream processors handle `FileBackedOp` transparently (already implemented)
- Also removes `wait_until_downstream_empty` from `on_terminate` since spill mode makes it unnecessary

This keeps bounded channels for backpressure during normal processing while preventing deadlocks during shutdown. Zero data loss.

## Files changed
- `engine/runtime/runtime/src/forwarder.rs` — SpillFile struct, spill_mode flag, `send_op_or_spill()`, `spill_feature()`, `flush_spill_files()`
- `engine/runtime/runtime/src/executor/processor_node.rs` — Enable spill mode before `finish()`, flush after

## Verified on GCP Batch

| Job | Data | Before | After | Spilled |
|-----|------|--------|-------|---------|
| Okayama fld QC | 2.3 GB | 6h timeout (deadlock) | **4.1h success** | 5,679 features |
| Okayama tran QC | 63 MB | 6h timeout (deadlock) | **6h timeout (no deadlock, genuine compute)** | 185,842 + 34,936 features |

The tran QC still times out because the `01-02-common` subgraph's XmlFragmenter takes 4+ hours on e2 shared vCPUs for this dataset — this is genuine computation, not a deadlock. The spill mechanism works correctly (zero `wait_until_downstream_empty` timeout cascade, all features delivered via spill files).

- Zero OOM errors (bounded channels preserved)
- Spill files automatically cleaned up with executor cache


engine/runtime/runtime/src/forwarder.rs
```diff
@@ -222,6 +254,38 @@ impl ChannelManager {
...
+                    ctx: ExecutorContext {
+                        port: port.clone(),
+                        ..ctx.clone()
+                    },
+                };
+                match sender.sender.try_send(op) {
+                    Ok(()) => {}
+                    Err(crossbeam::channel::TrySendError::Full(_)) => {
+                        // Use upstream port as spill key so flush can resolve mapping
+                        self.spill_feature(idx, &ctx.port, &ctx.feature);
```
n4to4: ## Potential data duplication when a sender has multiple port mappings from the same upstream port

In `send_op_or_spill` (`forwarder.rs:260-287`), the code iterates over each mapped downstream port and calls `try_send` individually. When some ports succeed and others hit a full channel, the feature is spilled to disk. However, the spill key is `(sender_idx, ctx.port)` — i.e., keyed by the **upstream** port, not the individual downstream port that failed.

Later, `flush_spill_files` calls `sender.send_file_backed_op(&dest, &port, context)`, which resolves the port mapping again and sends the `FileBackedOp` to **all** mapped downstream ports — including the ones where `try_send` already succeeded during the original `send_op_or_spill` call.

This means features that were already delivered inline to some ports will be delivered again via the spill file, resulting in **duplicate processing** downstream.

Additionally, within the inner `for port in ports` loop, if `try_send` fails on multiple ports for the same feature, `spill_feature` is called once per failure — writing the **same feature multiple times** to the spill file.

**Example scenario** (single sender, upstream port `"default"` mapped to downstream ports `["A", "B"]`):

1. `try_send(Op{port: A})` → **succeeds** (feature delivered to downstream)
2. `try_send(Op{port: B})` → **Full** → `spill_feature(idx, "default", feature)` (feature written to spill file)
3. `flush_spill_files` → `send_file_backed_op(file, "default", ctx)` → resolves mapping → sends `FileBackedOp` to **both A and B**
4. **Result**: Port A receives the feature **twice** (once inline, once via spill file)

**Possible fix**: Track which `(sender_idx, downstream_port)` pairs actually failed `try_send`, and only flush to those specific ports. Alternatively, if partial success within a single sender's port set is hard to track, spill the feature on the **first** failure for a sender and skip the remaining `try_send` calls for that sender entirely — then the spill file covers all ports for that sender consistently.

shunski (approved):

PR #2054, opened 2026-04-17
fix(engine): line on line overlayer performance improvements

shunski:
# Overview
This PR rewrites `LineOnLineOverlayer` with the R-tree candidate pair lookup and a careful disk-backing to avoid quadratic memory usage. It also contains some bug fixes for the QC tran workflow.

## What I've done
- `LineOnLineOverlayer` was using O(n^2) intersection checks for AABBs, which were making the action take more than 6h for the QC Okayama tran workflow. After the following major rewrite, now it completes in under 10 mins:
  - It now uses an R-tree for intersecting AABB pair lookup. Furthermore, since we iterate over the candidate pair instead of computing all AABB pairs upfront, O(n^2)-many pair candidates will not be loaded onto memory at once.
  - Disk-backing is introduced to prevent geometry data piles up in memory before `finish()`. (See the memo below for a trade-off made for this) 
- There is one behavioral change; `LineOnLineOverlayer` do not output lines with length less than tolerance (i.e. degenerate line strings). 
- This PR contains bug fixes for QC tran workflow (though they do not affect the inspection results, as all of them are related to an "OK error" 隣接面重複): 
  - It was using stale tolerance values based of the geographic coordinates from early days, though now it uses Cartesian local coordinates. These tolerance values are fixed.
## Memo
- `LineOnLineOverlayer` still loads all geometries onto memory at `finish()`. We could further modify it so that the each geometry will be loaded to memory only when it's one from a candidate AABB pair, but this PR does not implement it. This is a trade-off with memory and speed; we chose speed, and this is likely fine as the memory usage is still linear.

### DIFF
--- a/engine/runtime/action-processor/src/geometry/line_on_line_overlayer.rs
+++ b/engine/runtime/action-processor/src/geometry/line_on_line_overlayer.rs
@@ -427,11 +427,6 @@ impl DiskBackedFeatures {
         Ok(serde_json::from_slice(&buf)?)
     }
 
-    fn read_geometry(&self, i: usize) -> Result<Arc<Geometry>, BoxedError> {
-        let feature = self.read_feature(i)?;
-        Ok(feature.geometry)
-    }
-
     fn len(&self) -> usize {
         self.offsets.len()
     }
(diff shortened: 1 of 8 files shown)


engine/runtime/action-processor/src/geometry/line_on_line_overlayer.rs
```diff
@@ -193,354 +361,497 @@ impl Processor for LineOnLineOverlayer {
...
+        let mut included_feats: std::collections::HashSet<usize> =
+            std::collections::HashSet::from([feat_i]);
+
+        for cand in seg_rtree.locate_in_envelope_intersecting(&seg_aabbs[i]) {
+            let j = cand.idx;
+            if j <= i || processed[j] {
+                continue;
+            }
+            let (idx2, ls2) = (segments[j].0, &segments[j].1);
+            if segments_match(&ls1, ls2, tolerance) {
```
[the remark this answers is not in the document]
shunski: The tolerance is stated to be:
1. Criteria for identifying points, and
2. Detecting the degenerate line string output.
Two close parallel lines will not be detected by the tolerance.

engine/runtime/action-processor/src/geometry/line_on_line_overlayer.rs
```diff
@@ -193,354 +361,497 @@ impl Processor for LineOnLineOverlayer {
...
+        .collect::<Result<Vec<_>, _>>()?;
+
+    let lss_per_feature: Vec<Vec<LineString2D<f64>>> = geometries
+        .iter()
+        .map(|g| match &g.value {
+            GeometryValue::FlowGeometry2D(g2) => extract_line_strings(g2),
+            _ => Vec::new(),
+        })
+        .collect();
+
```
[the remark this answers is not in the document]
shunski: Added a top level count-check for indices.

engine/runtime/action-processor/src/geometry/line_on_line_overlayer.rs
```diff
@@ -193,354 +361,477 @@ impl Processor for LineOnLineOverlayer {
...
         }
+        Ok(Self {
+            path: path.to_path_buf(),
+            offsets,
+            lengths,
+        })
+    }
+
+    fn read_feature(&self, i: usize) -> Result<Feature, BoxedError> {
+        let mut file = File::open(&self.path)?;
```
asrcpq, line 427: Maybe not reopening file for each feature?

Also, I understand that current offsets+seek is for future optimization. Since currently a linear read seems to do the same job. In that case, maybe some comments in code helps understanding?

engine/runtime/examples/fixture/workflow/quality-check/plateau4/03-tran/workflow.yml
```diff
@@ -327,9 +327,9 @@ graphs:
           groupBy:
             - lod
             - fileIndex
-            - gmlId
+            - gmlRootId
```
asrcpq: Line 523 still reports with gmlId, but the dissolved feature can have constructed gmlId from a random child feature, so the reported gml_id may not be the relevant one. Maybe report the  root ID only? If the exact child feature's ID needs to be reported, it will become more complicated so I am not sure here.

engine/runtime/action-processor/src/geometry/line_on_line_overlayer.rs
shunski: You are right, and this design is due to possible performance tuning in the future (we might make it multi-threaded later). I have added some comments to the function explaining this.

engine/runtime/examples/fixture/workflow/quality-check/plateau4/03-tran/workflow.yml
shunski: This issue is a bit more subtle, and I decided to think more on this in another PR; for now, I have reverted the change.
(The fix is not urgent at all since this is one of those optional inspections which do not affect the overall inspection result. This is not part of the fix that was requested either. Anyway, I will come back to this later.)

asrcpq (approved):
LGTM

PR #2056, opened 2026-04-20
feat(engine): CityGML 3.0 child feature extraction

asrcpq:
## Overview

This PR implements `extractedTags` parameter in CityGML 3.0 reader. This is equivalent to `flatten=true` in `FeatureCitygmlReader`, but users must specify all tags that are considered a "feature", instead of relying on hard-coded PLATEAU-specific tag names. Some other options and problems:

1. parse XSDs of ADE: users still need to provide ADE files, or reader need to fetch HTTP XSD files which is unreliable
2. create another PLATEAU CityGML reader with feature type hard-coded: this leaves CityGML 3.0 reader not supporting ADE, which is probably not wanted.

## Changes

- feature CityGML3 reader: add extractedTags parameter to specify feature tags to be extracted as toplevels.
- create `__citygml_parent_gml_id` and `__citygml_root_gml_id` to preserve topology info.
- a subfeature xlink referred multiple times will be emitted multiple times with referrer's gml:id as parent ID.
- cesium 3D tiles writer: should not simply fail whole processor when geometry is missing or wrong geometry type in a single feature. Now logs a warning and skip.

## Notes

- parent/root ID are orthogonal to extractedTags. They should always point to the direct parent/toplevel feature in the original XML even they are not extracted.
- currently no stack overflow protection for recursive tree traversal, maybe add a depth limitation in the future

### DIFF
--- a/engine/runtime/action-processor/src/feature/reader/citygml3/flatten.rs
+++ b/engine/runtime/action-processor/src/feature/reader/citygml3/flatten.rs
@@ -1,0 +1,155 @@
+use std::collections::HashSet;
+use std::sync::Arc;
+
+use super::utils::{local_name, XmlChild, XmlNode};
+
+/// Walks `node`'s subtree bottom-up, lifting out every descendant whose tag is in `included`
+/// (matched by qualified name, local name, or Clark notation `{ns}local`).
+/// Returns the stripped tree and the extracted nodes, deepest-first.
+/// No-op when `included` is empty.
+pub fn extract_by_types(
+    node: &Arc<XmlNode>,
+    included: &HashSet<String>,
+) -> (Arc<XmlNode>, Vec<Arc<XmlNode>>) {
+    if included.is_empty() {
+        return (Arc::clone(node), Vec::new());
+    }
+    let mut extracted: Vec<Arc<XmlNode>> = Vec::new();
+    let stripped = extract_recursive(node, included, &mut extracted);
+    (stripped, extracted)
+}
+
+pub fn tag_matches(node: &XmlNode, included: &HashSet<String>) -> bool {
+    let ln = local_name(&node.name.0);
+    included.contains(node.name.0.as_str())
+        || included.contains(ln)
+        || (!node.name.1.is_empty()
+            && included.contains(&format!("{{{}}}{ln}", node.name.1)))
+}
+
+fn extract_recursive(
+    node: &Arc<XmlNode>,
+    included: &HashSet<String>,
+    out: &mut Vec<Arc<XmlNode>>,
+) -> Arc<XmlNode> {
+    let mut new_children: Option<Vec<XmlChild>> = None;
+
+    for (i, child) in node.children.iter().enumerate() {
+        match child {
+            XmlChild::Element(e) => {
+                if tag_matches(e, included) {
+                    // Process the child's own subtree first (bottom-up), then lift the child out.
+                    let stripped_child = extract_recursive(e, included, out);
+                    out.push(stripped_child);
+
+                    if new_children.is_none() {
+                        new_children = Some(node.children[..i].to_vec());
+                    }
+                    // deliberately not pushed into new_children — it is extracted
+                } else {
+                    let stripped_child = extract_recursive(e, included, out);
+                    match new_children {
+                        None => {
+                            if !Arc::ptr_eq(&stripped_child, e) {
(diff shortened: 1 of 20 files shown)


engine/runtime/action-processor/src/feature/reader/citygml3/processor.rs
```diff
@@ -176,20 +196,36 @@ impl Processor for FeatureCityGml3Reader {
...
+                    &ctx,
+                    feature,
+                    DEFAULT_PORT.clone(),
                 ));
+            } else {
+                let root_gml_id = gml_id_attr(&feature_root);
+                let mut tracker = ParentIdTracker::new();
+                tracker.collect(&feature_root);
+
+                for node in flatten::extract(&feature_root, &self.extracted_tags) {
```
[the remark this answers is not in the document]
asrcpq: if totally no feature was extracted, it should be easily observable from the intermediate data/frontend, so no need to add redundant log here.

engine/runtime/action-processor/src/feature/reader/citygml3/flatten.rs
```diff
@@ -0,0 +1,250 @@
...
+    let child_parent = my_id.as_deref().or(parent_gml_id);
+
+    let mut new_children: Option<Vec<XmlChild>> = None;
+
+    for (i, child) in node.children.iter().enumerate() {
+        match child {
+            XmlChild::Element(e) => {
+                if tag_matches(e, sets) {
+                    let stripped_child = extract_recursive(e, sets, out, child_parent);
+                    out.push((stripped_child, child_parent.map(str::to_string)));
```
n4to4, line 104: **Potential issue: listing both a parent tag and a descendant tag in `extractedTags` silently strips geometry from the parent**

In `flatten.rs:95-107` (`extract_recursive`), when a child element also matches `extractedTags`, it is emitted as a standalone feature and deliberately removed from the rebuilt parent:

```rust
if tag_matches(e, sets) {
    let stripped_child = extract_recursive(e, sets, out, child_parent);
    out.push((stripped_child, child_parent.map(str::to_string)));
    // deliberately not pushed into new_children — it is extracted
}
```

This interacts badly with CityGML 3's thematic-surface model. A `bldg:Building` typically carries its actual geometry through boundary surfaces like `con:FloorSurface` / `con:WallSurface` / `con:RoofSurface` — the `<bldg:Building>` element itself often has no direct geometry. If both are listed in `extractedTags`:

1. `bldg:Building` matches → `extract_recursive` walks its subtree.
2. `con:FloorSurface` matches → emitted as a standalone feature, **removed from the Building's children**.
3. The emitted Building has its geometry-bearing children stripped → effectively empty geometry.
4. Downstream, the new "skip on missing geometry" path added in this same PR (`sink.rs:192-207`) drops the Building with a warn log.
5. Net effect: building-level features silently disappear from the 3D tiles output; only the detached surfaces remain.

The example workflow added in this PR actually demonstrates this exact combination:

```yaml
# engine/runtime/examples/fixture/workflow/examples/citygml3-reader/workflow.yml
extractedTags:
  - bldg:Building
  - con:FloorSurface
  - tran:Road
  - tran:TrafficArea
  - frn:CityFurniture
```

A user copying this example would silently lose all building-level features.

**Options to consider:**
- (a) Keep the original subtree on the parent *and also* emit matching descendants as standalone features (duplicated emit, closer to `flatten=true` semantics in the CityGML 2 reader).
- (b) Keep the current stripping, but detect parent+descendant pairs in `extractedTags` and warn/error at processor startup.
- (c) Document the "outermost match wins / geometry moves to the child" rule explicitly, and revise the example to avoid overlapping pairs (e.g., pick only leaf surfaces, or only `bldg:Building`).

Whichever semantics is chosen, the current silent-drop-then-sink-skip chain combined with an example that triggers it seems likely to confuse users.

n4to4 (approved):
👍

engine/runtime/action-processor/src/feature/reader/citygml3/flatten.rs
asrcpq: The actual design is that child geometry should be excluded from parent feature. The example was intentionally constructed to test empty geometry get skipped without failing whole cesium writer, but it is probably a bit confusing.

Fix: update the GML to let some toplevel features to hold geometry, and add a comment to explicitly state child features' geometry extracted from parent.

PR #2067, opened 2026-04-24
fix(api): GCS setup issue [FLOW DEV 83]

KaWaite:
# Overview


pyshx (approved):
LGTM! technically we're still not calling `f.client.Close()` but that's okay since its a process-lifetime singleton. 

Not closing it is intentional, since the process exit reclaims it.

PR #2071, opened 2026-05-07
fix(engine): Add comment to suppress false positive traversal risk report

asrcpq:
# Overview

This PR adds a comment to #2066 in code to suppress AI security scanner reporting it as a vulnerability.


n4to4 (approved):
👍

PR #2075, opened 2026-05-07
feat(engine): CityGML 3.0 toplevel srsName/EPSG code parsing

asrcpq:
# Overview

This PR implements srsName parsing when specified in toplevel `boundedBy -> gml:Envelope` tag. This enabled a basic `CityGML3 reader -> reprojector -> cesium tiles writer` pipeline to work on a sample PLATEAU GML file.

## Notes

- GML allows per-geometry srsName overrides but considered rare in PLATEAU dataset so only implemented default toplevel ones (nusamai is somehow parsing per-geometry EPSG but still not storing them properly).
- currently hard match EPSG URL prefix. The final handling method depends on geometry refactoring (i.e., whether we should store EPSG code or srsName URLs in geometry traits)


asrcpq:
Closing as EPSG is currently not a blocker of parsing PLATEAU citygml 3.0 files. May reopen in the future when needed.

PR #2094, opened 2026-05-18
fix(api): adjust permission check for API-driven triggers

KaWaite:
# Overview


pyshx (approved):
Makes Sense! Thanks for this change.

PR #2101, opened 2026-05-19
feat(ui): add user focused element handling in various components [FLOW-DEV-133]

billcookie:
# Overview
Adds a `focusedElement` field to the Yjs awareness state so collaborators' cursors can be suppressed while they're interacting with modal/popover UI (action picker, params dialog, layout options, debug popovers, search panel, homebar dialogs). A new `handleUserFocusedElement` awareness setter is threaded through the Editor, OverlayUI, Homebar, DebugActionBar, ParamsDialog and `useUIState` hook, and `MultiCursor` hides remote cursors when the remote user has `focusedElement` set.

The general trend with the PR is that essentially if a user is either in a dialog or a popover then their cursor will no longer show for other users.
## What I've done
- Add `focusedElement?: boolean` to `AwarenessUser` and a `handleUserFocusedElement` setter in `useAwarenessPresence`.
- Propagate `onUserFocusedElement` callbacks through Editor → OverlayUI / ParamsDialog / Homebar / DebugActionBar / `useUIState`, and call it from open/close handlers (including a new `handleShowSearchPanel` wrapper and refactor of `LayoutOptionsDialog` to use the shared `DialogOptions` state).
- `MultiCursor` returns null when the remote `user.focusedElement` is true.
## Memo

### DIFF
--- a/ui/src/features/Editor/components/OverlayUI/components/LayoutOptionsDialog/index.tsx
+++ b/ui/src/features/Editor/components/OverlayUI/components/LayoutOptionsDialog/index.tsx
@@ -18,10 +18,7 @@ import {
 import { useT } from "@flow/lib/i18n";
 import type { Algorithm, Direction } from "@flow/types";
 
-import { DialogOptions } from "../../types";
-
 type Props = {
-  showDialog: DialogOptions;
   onClose: () => void;
   onLayoutChange: (
     algorithm: Algorithm,
@@ -30,11 +27,7 @@ type Props = {
   ) => void;
 };
 
-const LayoutOptionsDialog: React.FC<Props> = ({
-  showDialog,
-  onClose,
-  onLayoutChange,
-}) => {
+const LayoutOptionsDialog: React.FC<Props> = ({ onClose, onLayoutChange }) => {
   const t = useT();
 
   const [layoutDirection, setLayoutDirection] =
@@ -65,7 +58,7 @@ const LayoutOptionsDialog: React.FC<Props> = ({
   };
 
   return (
-    <Dialog open={!!showDialog} onOpenChange={(o) => !o && onClose()}>
+    <Dialog open={true} onOpenChange={(o) => !o && onClose()}>
       <DialogContent size="sm" position="top" overlayBgClass="bg-opacity-0">
         <DialogTitle>{t("Layout Options")}</DialogTitle>
         <DialogContentWrapper>
--- a/ui/src/features/Editor/components/OverlayUI/components/LayoutOptionsDialog/index.tsx
+++ b/ui/src/features/Editor/components/OverlayUI/components/LayoutOptionsDialog/index.tsx
@@ -18,8 +18,10 @@ import {
 import { useT } from "@flow/lib/i18n";
 import type { Algorithm, Direction } from "@flow/types";
 
+import { DialogOptions } from "../../types";
+
 type Props = {
-  isOpen: boolean;
+  showDialog: DialogOptions;
   onClose: () => void;
   onLayoutChange: (
     algorithm: Algorithm,
@@ -29,7 +31,7 @@ type Props = {
 };
 
 const LayoutOptionsDialog: React.FC<Props> = ({
-  isOpen,
+  showDialog,
   onClose,
   onLayoutChange,
 }) => {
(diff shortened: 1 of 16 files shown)


ui/src/features/Editor/hooks.ts
```diff
@@ -322,8 +315,8 @@ export default ({
           if (hasModifier) handleProjectLockChange?.(!isLocked);
           break;
         case "k":
-          if (hasModifier && !showSearchPanel) setShowSearchPanel(true);
-          if (hasModifier && showSearchPanel) setShowSearchPanel(false);
+          if (hasModifier && !showSearchPanel) handleShowSearchPanel(true);
+          if (hasModifier && showSearchPanel) handleShowSearchPanel(false);
```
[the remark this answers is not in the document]
billcookie: ✅

ui/src/features/Editor/components/ParamsDialog/index.tsx
```diff
@@ -134,6 +136,10 @@ const ParamsDialog: React.FC<Props> = ({
     [setMyDraft],
   );
 
+  useEffect(() => {
+    onUserFocusedElement?.(!!openNode);
+  }, [openNode, onUserFocusedElement]);
```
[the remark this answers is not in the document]
billcookie: ✅

ui/src/features/Editor/components/OverlayUI/components/LayoutOptionsDialog/index.tsx
```diff
@@ -63,7 +65,7 @@ const LayoutOptionsDialog: React.FC<Props> = ({
   };
 
   return (
-    <Dialog open={isOpen} onOpenChange={(o) => !o && onClose()}>
+    <Dialog open={!!showDialog} onOpenChange={(o) => !o && onClose()}>
```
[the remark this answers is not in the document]
billcookie: ✅

ui/src/lib/yjs/useAwarenessPresence.ts
```diff
@@ -179,6 +179,13 @@ export default function useAwarenessPresence({
     [yAwareness],
   );
 
+  const handleUserFocusedElement = useCallback(
+    (isOpen: boolean) => {
+      yAwareness.setLocalStateField("focusedElement", isOpen);
```
KaWaite, line 184: Doesn't have to be in this PR, but it would be nice to stop tracking cursor information while the user is focused. Pretty sure the current implementation merely hides the cursor UI, but will still be updating the yAwareness, which is unneeded activity.

ui/src/features/Editor/components/ParamsDialog/index.tsx
```diff
@@ -134,6 +136,15 @@ const ParamsDialog: React.FC<Props> = ({
...
   );
 
+  const onUserFocusedElementRef = useRef(onUserFocusedElement);
+  onUserFocusedElementRef.current = onUserFocusedElement;
+  useEffect(() => {
+    onUserFocusedElementRef.current?.(true);
+    return () => {
+      onUserFocusedElementRef.current?.(false);
+    };
+  }, []);
```
KaWaite: This code seems pointless. Is there any purpose of this?

billcookie: Something like this would be better:

```
useEffect(() => {
    onUserFocusedElement?.(true);
    return () => onUserFocusedElement?.(false);
  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```

KaWaite: Ah, sorry now I understand what is happening in the code. Was reading it wrong.
I still find this strange as all other dialogs have the `focusedElement` set to true on open, and set to false on close.
Why does only the `ParamsDialog` need to set `focusedElement` inside the dialog in this way rather than adding the change into the dialog open setter like all other dialogs?

ui/src/lib/yjs/useAwarenessPresence.ts
```diff
@@ -179,8 +179,16 @@ export default function useAwarenessPresence({
...
+    },
+    [yAwareness],
+  );
+
   useEffect(() => {
     const handleWindowPointerMove = (event: PointerEvent) => {
+      if (yAwareness.getLocalState()?.focusedElement) return;
```
KaWaite, line 191: I am pretty sure this will be stuck to whatever focusedElement is each time a dep changes and forces a run of the useEffect.
Since yAwareness is a ref, I think this means that this won't get updated when focusedElement changes. But, I admit due to it being refs + addEventListeners, I could be wrong.
Can you do some checks to make sure pointer move is actually returning here and not still passing through when focsedElement becomes true?

KaWaite (approved):

PR #2107, opened 2026-05-21
fix(api): finalize Cloud Run debug jobs via the monitoring loop

pyshx:
# Overview

Follow-up to #2103. Cloud Run debug jobs (editor "Run") never updated their status, and **Output data** stayed empty, even when the workflow ran successfully.

## What I've done

Root cause: `RunDebugJob` persisted the final status with `tx.Context()` — i.e. *inside* a MongoDB transaction — while every other status path (`checkJobStatus`, `Cancel`, `saveJobState`) saves with the plain `ctx`. That transaction routinely aborts in this environment (`transaction end failed: NoSuchTransaction ... has been aborted`), which is harmless for the other paths (their writes auto-commit) but rolled back `RunDebugJob`'s status write. The same failed finalization is why `outputURLs` were never set ("Output data (0)").

- Deleted `RunDebugJob`'s bespoke poll-and-finalize logic.
- Added `RunCloudRunWorker`, which triggers execution on the worker Service and finalizes **only** on infra failure (where no Redis worker event will arrive).
- `project.go` debug branch now calls `RunCloudRunWorker` **and** `StartMonitoring` — the same loop Batch uses. `checkJobStatus` already handles debug jobs (empty `GCPJobID` skips Batch polling; the builder leaves `batchStatus` nil so the status merge returns the Redis worker status), and it persists status + runs `handleJobCompletion` (which sets `outputURLs`).
- Removed the debug-job skip in `startMonitoringIfNeeded`.

Net −40 lines; success/workflow-failure finalize via the proven loop, infra failures via `failCloudRunJob`.

## What I haven't done

- No change to `workerArtifactPath` — the worker's output path (`artifacts/<jobID>/artifacts/`) already matches `ListJobArtifacts`.
- Intermediate-data compression (`FLOW_RUNTIME_COMPRESS_INTERMEDIATE_DATA=true`) is a worker-Service env setting (already applied on the deployed Service); codifying it in terraform is tracked with the rest of the debug-Service infra.

## How I tested

- `go build ./...`, `go vet`, `go test ./pkg/job/... ./internal/usecase/interactor/...` — all pass.
- Verified in reearth-oss logs that the worker runs and publishes the `JobCompleteEvent` to Redis, and that `checkJobStatus` reads it on the debug (empty `GCPJobID`) path.

## Screenshot

n/a

The reliance on `batchStatus` being nil for debug jobs (so the merge returns the worker status), and the split between `RunCloudRunWorker` (infra failure) and the monitoring loop (success / workflow failure).

## Memo

Follow-up to #2103. Needs the worker Service env `FLOW_RUNTIME_COMPRESS_INTERMEDIATE_DATA=true` (already set) for intermediate Data Preview.

### DIFF
--- a/server/api/internal/usecase/interactor/job.go
+++ b/server/api/internal/usecase/interactor/job.go
@@ -156,106 +156,65 @@ func (i *Job) Cancel(ctx context.Context, jobID id.JobID) (*job.Job, error) {
 	return j, nil
 }
 
-// RunDebugJob runs a Cloud Run debug job end-to-end: blocks on RunJob, then
-// waits (bounded) for the authoritative JobCompleteEvent from Redis before
-// persisting the final merged status and calling handleJobCompletion.
-func (i *Job) RunDebugJob(ctx context.Context, j *job.Job, p gateway.RunJobParam) {
-	// This runs in a detached goroutine; a panic here would otherwise crash the API server.
-	defer func() {
-		if r := recover(); r != nil {
-			log.Errorfc(ctx, "job: debug job %s panicked: %v", j.ID(), r)
+// RunCloudRunWorker triggers execution of a debug job on the Cloud Run worker
+// Service. The worker publishes its authoritative result to Redis (via the
+// subscriber); the standard monitoring loop reads that event and finalizes the
+// job, exactly as it does for Batch jobs. Only an infrastructure failure — where
+// no result will ever arrive — is finalized here.
+func (i *Job) RunCloudRunWorker(j *job.Job, p gateway.RunJobParam) {
+	go func() {
+		ctx := context.Background()
+		// Detached goroutine; a panic here would otherwise crash the API server.
+		defer func() {
+			if r := recover(); r != nil {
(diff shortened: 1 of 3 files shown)


pyshx:
Thanks @copilot — both addressed in 69be56a5a:

1. **Premature/false failure on a client-side error** — `failCloudRunJob` now waits a bounded grace period (`cloudRunInfraFailureGrace`, 30s) for a late worker completion event, and bails out if the job is already terminal or a `JobCompleteEvent` has since landed in Redis (the monitoring loop finalizes from it). It also records the failure via the legacy `status` field, leaving `batchStatus` nil, so a later `workerStatus=COMPLETED` is no longer forced to `FAILED` by the AND/OR merge.

2. **Duplicate completion side-effects** — `checkJobStatus` now gates `handleJobCompletion` on `statusChanged && isTerminalState`, so completion handling only runs on an actual transition. If `failCloudRunJob` (or `Cancel`) already finalized the job, the loop won't re-fire notifications/artifact listing.

PR #2109, opened 2026-05-25
feat(engine): Migrate AttributeManager to FlowExpr

asrcpq:
# Overview

Continue #2030 and #2087, migrate AttributeManager.

## Changes

- migrate AttributeManager and related workflows to FlowExpr.
- support bit operations on positive integer.
- support scientific notation in float parsing.
- support `module` value type. Implement basic math module.
- migrate `.len()` to global `len()` to align with Python convention.
- change attribute/env value access to indexing style `attributes[key]` and `attributes.get(key[, default])`. Support attribute key iteration (env does not support iteration).
- fix AttributeManager's rename field to evaluate code. Previously Expr type was used but the code was used as string literals.
- url.stem is removed. It is not common to have a stem method in Url library so need further discussion.

## Known breaking changes

- some `value(key)` are migrated to `attributes[key]` instead of `attributes.get(key)` for early data corruption detection.
- dc/01-bldg's highest lod extraction now gives -1 instead of 0 if `__citygml_lod_mask` is zero - currently impossible to hit so should be safe.
- NaN/Inf will raise error when converted back to Flow value (previously silently fallback to null). solar-radiation related workflows may fail for bad input data such as zero division or negative log.

## Misc

- known limitation (pre-existing) in AttributeManager: Rename conflict detection checks the original input feature for new_key, not the accumulated result. If two Rename operations target the same new_key, the second silently overwrites the first because neither finds new_key in the original feature.
- round() is implemented under math module and use away-from-zero semantics.
- for simplicity, nested complex types are deeply cloned which may be a problem for some huge attribute value.
- most workflow translations follow original code structure to reduce possible compatibility breaks. Therefore the migrated version may not be the most idiomatic FlowExpr implementation.
- todo: evaluate error handling is inconsistent (pre-existing). Should add a new error type for recoverable process() error. So they can be suppressed in production with error log while triggering test/CI failure.


n4to4 (approved):
👍

PR #2114, opened 2026-05-26
refactor(engine/api): Added hidden and tags to actions and improved categorization  [FLOW DEV 98]

KaWaite:
# Overview

This update introduces a tagging system for various factories and actions as well as revamped categories to improve organization and discoverability.
Also, added a hidden attribute to start showing to the UI only the base-suite of actions.

## What I've done

- Added tags to SourceFactory, ProcessorFactory, and SinkFactory traits.

- Enhanced action schema with tags and hidden properties.

- Updated categories for several factories to better reflect their functionality.

- Implemented filtering for hidden actions and included tags in action summaries.

## What I haven't done

- No additional features outside of tagging and categorization enhancements.

## How I tested

- Verified the functionality of tags and categories through unit tests.

## Screenshot

- Not applicable.

- Review the consistency and relevance of the tags added to ensure they meet the intended categorization goals.

## Memo

- This update addresses the need for improved organization within the system.


n4to4 (approved):
I’d prefer to define tag as an enum rather than a string, but category is already defined as a string, so I think it would be fine to refactor both together later.

(I’d rather avoid using strings too much, though.)

PR #2123, opened 2026-06-01
feat(engine/sink)!: require relative paths in sink output

### LINKED ISSUE #2117: feat(engine/sink): sinks should accept relative paths and join internally (follow-up to #2108)
**Is your feature request related to a problem? Please describe.**

PR #2108 ("sandbox sink writes against per-job artifact directory") delivered the **security half** of the original task — every sink write is now bound to the per-job artifact root via `SinkOutput::from_path` → `sandbox::ensure_under`. The **ergonomics half** of the original ask was not delivered.

The original task description had two parts:

> Instead of providing a environment (like plateau's `workerArtifactDir`) and manually concatenate full URL in every sink action, sink actions should give **relative path** (such as `foo/bar.txt`, traversal protection needed after concat). The full Url construction should be handled internally.

| Goal | Status |
|---|---|
| Traversal protection / bounded writes | ✅ Shipped in PR #2108 |
| Sinks accept **relative paths**, engine concats internally | ❌ Still missing |

Today, every workflow still has to manually concatenate the artifact path:

```yaml
# Current pattern (post-PR2)
output:
  type: flowExpr
  value: |
    Url(env("workerArtifactPath")) / "bldg_lod1"
```

77+ existing plateau fixtures, plus any customer workflows, still use this pattern. Workflow authors must know about `workerArtifactPath` and remember to call `Url(env(...))/...`.

**Describe the solution you'd like**

Workflows should be able to give sinks a **relative path**, and the engine should handle the join internally against `ctx.sandbox_root`:

```yaml
# Proposed pattern
output: "bldg_lod1"
```

Concretely:

1. **Sink param semantics change** — `output`, `compress_output`, and equivalents accept either:
   - A bare relative path (`"bldg_lod1"` / `"group/a.geojson"`)
   - A full URI (`"gs://bucket/x"` / `"file:///abs/path"`) — for cases where a workflow truly needs to write to an explicit absolute location (rare; should still be sandboxed)

2. **Engine-side join** — `SinkOutput::from_path(ctx, path)` detects whether `path` parses as an absolute URI:
   - Absolute URI → use as-is (still sandbox-validated)
   - Relative path → join against `ctx.sandbox_root`, then validate

3. **Deprecate `workerArtifactPath` global** — once relative paths are the norm, the workflow author no longer needs to know this env var exists. The worker/CLI still resolves it internally as `sandbox_root`.

**Describe alternatives you've considered**

| Approach | Pros | Cons |
|---|---|---|
| **A. Transparent dual-mode** (current proposal) | No breaking change; existing fixtures keep working; new workflows can use the simpler form | Two ways to do the same thing; ambiguity if a path could be parsed either way |
| **B. Hard breaking switch** | Single canonical form; cleaner API surface | All 77+ fixtures + customer workflows must migrate in lockstep; risk of breaking unknown users |
| **C. New explicit param shape** (e.g. `outputRelative` vs. `outputAbsolute`) | Unambiguous | Verbose; doesn't actually simplify ergonomics |
| **D. Do nothing** | Zero risk | Original task remains half-done; ergonomics gap persists |

Recommendation: **A** (transparent dual-mode) with a deprecation warning on the absolute-URI form. Migrate fixtures opportunistically. Drop the absolute form in a later major version if desired.

**Scope of work**

- [ ] Decide on migration strategy (A / B / C / D) — design discussion first
- [ ] `SinkOutput::from_path` detects relative vs. absolute, joins relative paths against `ctx.sandbox_root` before validation
- [ ] Same treatment for `SinkOutput::join` (currently relative-only against the captured root)
- [ ] Audit all sink param sites that take a path expression — confirm none have semantics that break under the new rules
- [ ] Update workflow author documentation to show the new form as the recommended pattern
- [ ] (Optional) Migrate the 77 plateau fixtures from `Url(env("workerArtifactPath"))/x` to `"x"`
- [ ] (Optional) Add a deprecation warning if the absolute form is used in a context where relative would work
- [ ] (Optional) Consider deprecating the `workerArtifactPath` workflow global if direct workflow access is no longer needed

**Acceptance criteria**

- Workflow YAML can specify `output: "foo/bar.txt"` for any sink and have it write to `<sandbox_root>/foo/bar.txt`
- Existing workflows using `Url(env("workerArtifactPath"))/x` continue to work without modification
- Sandbox check still rejects any path that would escape `sandbox_root`, regardless of whether it was authored as relative or absolute
- All existing sink integration tests pass
- A new test exercises the relative-path path end-to-end through at least one sink (e.g. GeoPackage or JSON)

**Additional context**

- Original task description (internal): the relative-path API was framed as the primary user-facing change. The sandbox was a security implication of the ask, not the headline ask
- PR #2108 (merged): https://github.com/reearth/reearth-flow/pull/2108
- The chokepoint this work attaches to: `engine/runtime/action-sink/src/output.rs` (`SinkOutput::from_path`, `SinkOutput::join`)
- The sandbox primitive: `engine/runtime/action-sink/src/sandbox.rs` (`ensure_under`)
- Affected fixture pattern (sample): `engine/runtime/examples/fixture/workflow/data-convert/plateau4/01-bldg.yml` and 76 sibling files
- Note: with the sandbox in place, this is purely an ergonomics / API simplification change. There is no remaining security exposure that this PR would close

bnimit:
## Summary

Sink `output` parameters now require a **relative path**. The engine joins it internally against the per-job artifact directory (`sandbox_root`). Workflow authors no longer need to write `Url(env["workerArtifactPath"]) / x`.

**This is a breaking change** for any workflow author using the old absolute-URI pattern. See the CHANGELOG entry for the migration steps.

Closes #2117.
Follow-up to #2108 (which delivered the security half — the sandbox check).

## What changed

- `SinkOutput::from_path` rewritten to validate + join against `ctx.sandbox_root`
- 16+ new unit tests covering the new validation chain (empty, whitespace, `.`/`..`, `://`, leading `/`/`~`, traversal, resolves-to-root, migration-keyword)
- 2 new integration tests (relative-path success + absolute-path failure with migration hint)
- 87 fixture YAML files migrated from legacy `Url(env["workerArtifactPath"])/x` and `file::join_path(env.get("workerArtifactPath"), x)` patterns
- FeatureWriter processor updated to use `SinkOutput::from_path` (was using `Uri::from_str` directly)
- 4 test harnesses switched from `Runner::run` (unsandboxed) to `Runner::run_with_sandbox_root(tempdir())` to support relative paths in tests
- `engine/AGENTS.md` "Sandbox & sink writes" section updated
- CHANGELOG entry with migration instructions
- Workspace version bumped

## Migration for workflow authors

Old:
```yaml
output:
  type: flowExpr
  value: |
    Url(env["workerArtifactPath"]) / "out.gpkg"
```

New (static):
```yaml
output:
  type: string
  value: "out.gpkg"
```

New (dynamic):
```yaml
output:
  type: flowExpr
  value: |
    "out_" + attributes["k"] + ".gpkg"
```

Workflows still using the old pattern fail at runtime with an error mentioning `workerArtifactPath` — search logs for that keyword.

## Test plan

- [x] `cargo make test-rs` green (945 passed, 0 failed)
- [x] `cargo make test-qc` green (135 passed, 0 failed)
- [x] `cargo make clippy` clean
- [x] `cargo make format` no-op (fmt committed)
- [x] Manual: grep `env["workerArtifactPath"]` and `env.get("workerArtifactPath")` across YAML fixtures — zero sink-output uses remain


engine/runtime/action-processor/src/feature/writer.rs
```diff
@@ -255,8 +255,15 @@ impl Processor for FeatureWriter {
...
-        let buffer = self.buffer.entry(output).or_default();
+        // Use SinkOutput::from_path so the path is validated as strict-relative
+        // and joined against sandbox_root, rather than against CWD.
+        let sink_out = reearth_flow_action_sink::SinkOutput::from_path(
+            &NodeContext::from(ctx.clone()),
+            path.as_str(),
+        )
+        .map_err(|e| FeatureProcessorError::FeatureWriterFactory(format!("{e}")))?;
+        let output_uri = sink_out.uri().clone();
+        let buffer = self.buffer.entry(output_uri).or_default();
```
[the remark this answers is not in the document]
bnimit: Acknowledged — same concern flagged by an internal code review during Task 3. The current shape is intentional but the perf concern is real.

**Trade-off:**
- `StorageResolver::resolve` caches by backend URL (`RwLock::read` + `Arc::clone` on hit), so the second feature onward is cheap
- The wider redesign would introduce a separate `validate_relative_path` helper that returns `Uri` without acquiring Storage, and would need to be applied consistently to `FeatureWriter`, cesium3dtiles `process_default`, and mvt `process` (all introduced or modified in this PR)

**Decision for this PR:** keep the per-feature `SinkOutput::from_path` call. The validation behavior is correct and the storage-resolver cache makes the hot path acceptable. A separate follow-up can extract the lighter helper if profiling shows it matters in practice.

Tracking as a follow-up consideration for a future "sink chokepoint hardening" PR alongside the related concern about helper functions (`write_citygml_to_storage`, `write_json`, `write_csv`) that trust callers to validate.

engine/runtime/examples/fixture/graphs/plateau4/quality-check/01-02-common.yml
```diff
@@ -87,7 +87,7 @@ nodes:
           feature["dataFileData"]
         }
       output: |
-        file::join_path(env.get("workerArtifactPath"), "summary_common2-" + env.get("__value")["package"] + ".json")
+        "summary_common2-" + env.get("__value")["package"] + ".json"
```
KaWaite, line 90: I believe if this is FlowExpr, output needs to be updated to have codeType = FlowExpr to work. Would like @asrcpq to confirm though

asrcpq: Should be okay since output in other writers are still using Rhai script.

engine/runtime/action-sink/src/file/citygml.rs
```diff
@@ -74,12 +74,28 @@ pub fn write_citygml_to_storage(
...
 
-    // Parent directory portion of the output URI string
-    let output_str = output.to_string();
-    let parent_str = output_str
+    // Compute the GML output's relative path under sandbox_root by stripping the
+    // sandbox_root prefix. This is used to derive the texture dst relative path.
+    let sandbox_root_str = sandbox_root.as_str().trim_end_matches('/');
+    let output_str = output.as_str();
+    // Strip the sandbox root prefix to get the GML's relative path (e.g. "group/foo.gml")
+    // Strip sandbox_root prefix to get the GML's relative path (e.g. "group/foo.gml")
```
asrcpq: comment repeated

engine/AGENTS.md
```diff
@@ -61,25 +61,56 @@ Use the `add-action` skill for a full step-by-step guide including i18n workflow
...
+```yaml
+output:
+  type: string
+  value: "out.gpkg"           # or "group/a.geojson", with attribute concat for dynamic names
+```
+
+The engine joins this against whatever `sandbox_root` the worker/CLI was
+launched with — `file:///var/jobs/abc/`, `gs://bucket/jobs/abc/`,
+`ram:///jobs/abc/`, etc. The same workflow YAML is portable across storage
+backends.
```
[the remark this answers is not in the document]
bnimit: Good catch — corrected the wording in `b38f1767f`. The chokepoint (`SinkOutput::new` plus the storage-resolver layer) IS scheme-agnostic, but multiple sinks bypass the storage abstraction with direct `std::fs::*` calls and effectively require a `file://` root:

- `ShapefileWriter` — `shapefile::Writer::from_path` + `std::fs::create_dir_all` / `remove_file`
- `GltfWriter` — `std::fs::create_dir_all` for atlas dir
- `Cesium3DTilesWriter` — atlas `create_dir_all` + post-zip `remove_dir_all`
- `MvtWriter` — post-zip `remove_dir_all`
- `GeoPackageWriter` — temp-file staging via `std::fs::read`

The updated AGENTS.md section now splits portability into two layers: the chokepoint is portable; individual sink portability varies and is enumerated. Closing the remaining `std::fs` gap (wrapping the rest behind the storage backend) is tracked as a future hardening — it's out of scope for #2117 (which is the relative-paths API change) but worth its own follow-up.

engine/runtime/action-sink/src/file/cesium3dtiles/pipeline.rs
```diff
@@ -373,9 +373,22 @@ pub(super) fn tile_writing_stage(
...
+    // We use it as the base URI for sub-paths written by this pipeline. Since we cannot
+    // pass it back through SinkOutput::new (which expects relative paths), we derive
+    // the tile sub-paths by stripping the sandbox_root prefix and composing strings.
+    let sandbox_root_str = node_ctx.sandbox_root.as_str().trim_end_matches('/');
+    let output_rel = output_path
+        .as_str()
+        .strip_prefix(sandbox_root_str)
+        .map(|s| s.trim_start_matches('/').to_string())
+        .unwrap_or_else(|| {
+            output_path
```
asrcpq: using flat filename as a silent fallback can cause data corruption. Invalid relative path should be fatal and should be a hard return Err instead. Similar problem for citygml.rs.

asrcpq (approved):
Approved with minor issues.

For filtering, there might be other corner cases like NUL character, URL vs path compatibility, etc. But since this PR's original motivation was simplification instead of sandboxing, it should be sufficient. We probably need to have a safe_join for Url class if everything is fixed to be url-based.

engine/runtime/action-sink/src/file/citygml.rs
bnimit: Fixed in `eeba0e298`. Removed the duplicate "Strip sandbox_root prefix..." comment line.

engine/runtime/action-sink/src/file/cesium3dtiles/pipeline.rs
bnimit: Good catch — that was a real correctness bug. Both `cesium3dtiles/pipeline.rs:386` and `citygml.rs` (same pattern) were silently falling back to `file_name()` when the `strip_prefix` failed. As you noted, that would collide across groups and corrupt data (e.g. `group_a/foo_appearance/bar.png` and `group_b/foo_appearance/bar.png` both writing to `foo_appearance/bar.png`).

Fixed in `eeba0e298`: both sites now return a hard `SinkError` if the strip fails. Since `output_path` is produced by `SinkOutput::new` (which is `sandbox_root.join(relative_path)`), the prefix MUST be present — if it ever isn't, something upstream is broken and we want it to surface loudly instead of producing wrong files.

asrcpq:
Sorry, mistakenly merged some other commits into it, reverted.

engine/runtime/examples/fixture/workflow/quality-check/plateau4/06-fld/workflow.yml
```diff
@@ -345,7 +344,6 @@ graphs:
         with:
           format: csv
           output: |
-            let base_path = env.get("workerArtifactPath");
             let folder = env.get("__value").Folder;
             if folder == () || folder == "" {
               return file::join_path(base_path, "06_浸水想定区域_fld.csv");
```
asrcpq, line 349: Regression. Some constructed paths are not updated. Here base_path is already removed so this branch will trigger error when hit.

PR #2126, opened 2026-06-02
refactor(engine): code language allowlist in schema definition

asrcpq:
# Overview

Currently Code in schema is a fixed type, so all code fields must have identical allowed language list.

This PR adds const generics to define accepted code types optionally for some fields. For example, conditions in FeatureFilter should not take plain string literal types. This PR also updates schema to pass the accepted types to the frontend.

## Changes

- added repr(u32) bitmask operations to CodeType.
- made Code generic over const MASK: u32 (defaulting to ALL_CODE_TYPES) to restrict accepted code types per field.
- manually implemented Serialize, Deserialize, and JsonSchema for Code<MASK> to enforce and expose the allowlist.
- inlined Code schema per usage site (instead of a shared $ref) since the schema varies by MASK.
- narrowed FeatureFilter's expr field to Code<{ CodeType::FlowExpr as u32 }>, excluding plain string literals.
- removed the impl Expr { compile() } method that was dead code.

### DIFF
--- a/engine/runtime/action-processor/src/feature/filter.rs
+++ b/engine/runtime/action-processor/src/feature/filter.rs
@@ -8,7 +8,7 @@ use reearth_flow_runtime::{
     forwarder::ProcessorChannelForwarder,
     node::{Port, Processor, ProcessorFactory, DEFAULT_PORT},
 };
-use reearth_flow_types::{Code, CompiledCode};
+use reearth_flow_types::{Code, CodeType, CompiledCode};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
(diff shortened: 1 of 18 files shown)


n4to4 (approved):
👍

engine/runtime/action-processor/src/feature/filter.rs
```diff
@@ -123,7 +123,7 @@ struct FeatureFilterParam {
 #[serde(rename_all = "camelCase")]
 struct Condition {
     /// # Condition expression
-    expr: Code,
+    expr: Code<{ CodeType::FlowExpr as u32 }>,
```
asrcpq, line 126: TODO: rebase and check all fields to determine what fields are code-only.

PR #2143, opened 2026-06-08
fix(engine): fix PLATEAU workflow migration #2123

asrcpq:
# Overview

This PR completes PLATEAU workflow migration for relative path output requirement introduced in #2123.

PLATEAU3 and requirements/PLATEAU4 workflows are deprecated so not touched.


asrcpq:
The fix will be handled together with #2142. Closing.

PR #2152, opened 2026-06-11
fix(ui): prevent backspace deletion for nodes/edges when project is locked [FLOW-DEV-189]

billcookie:
# Overview
This PR disables ReactFlow’s keyboard-driven element deletion when the editor is in a locked (read-only) state, preventing nodes/edges from being removed via Backspace/Delete while a project is locked.
## What I've done
- Set `deleteKeyCode` to `null` when `isLocked`/`readonly` to disable ReactFlow’s delete hotkey handling in read-only mode.

### DIFF
--- a/ui/src/features/Canvas/index.tsx
+++ b/ui/src/features/Canvas/index.tsx
@@ -147,6 +147,7 @@ const Canvas: React.FC<Props> = ({
       nodesConnectable={!readonly}
       nodesDraggable={!readonly}
       nodesFocusable={!readonly}
+      deleteKeyCode={readonly ? null : undefined}
       // elementsSelectable={!readonly}
 
       reconnectRadius={!readonly ? 10 : 0}


ui/src/features/Canvas/index.tsx
```diff
@@ -147,6 +147,7 @@ const Canvas: React.FC<Props> = ({
       nodesConnectable={!readonly}
       nodesDraggable={!readonly}
       nodesFocusable={!readonly}
+      deleteKeyCode={readonly ? null : undefined}
```
KaWaite, line 150: Can't... you just do the same as the others `!readonly`? 🤔

KaWaite (approved):

ui/src/features/Canvas/index.tsx
billcookie: There is a type issue if we do that: 

`Type 'boolean' is not assignable to type 'KeyCode | null | undefined'.`

PR #2173, opened 2026-06-17
fix(engine): iterative DAG traversal to prevent stack overflow on long workflows

bnimit:
## Summary

Long workflows could panic the engine with a **stack overflow** while *building* the executor. The cause was recursive traversal of the workflow/DAG whose recursion depth grows with workflow size. This converts the two such traversals to iterative algorithms — fixing the crash and, as a bonus, removing an O(N²) cost.

- **`DagSchemas::collect_ancestor_sources`** — a per-node recursive ancestor walk, called once per node by `BuilderDag::new`, so recursion depth = ancestor-chain length (overflow on long workflows) and total cost was O(N²). Replaced by a single **`affecting_sources_per_node()`** pass using `petgraph::algo::toposort` + forward propagation in topological order → constant stack depth, O(V+E). (Mirrors the existing `schema_infer::infer_and_validate` pattern.)
- **`collect_all_in_graph_recursive`** (downward subgraph-nesting walk in `incremental.rs`) → an explicit-stack worklist.

**Behavior is preserved exactly.** The new source-set pass is asserted identical to the old recursion via a differential test against a reference recursive oracle — including the subtle `kind == None` **barrier** semantics (a `None`-kind predecessor and its ancestors are excluded). A cyclic graph now returns a clean `ExecutionError::SchemaInferenceCycle` instead of recursing forever.

## Changes

- `engine/runtime/runtime/src/dag_schemas.rs` — add `affecting_sources_per_node`; remove `collect_ancestor_sources` + its recursive helper.
- `engine/runtime/runtime/src/builder_dag.rs` — call the batch method (one line).
- `engine/runtime/runtime/src/incremental.rs` — recursion → explicit-stack worklist.
- `engine/Cargo.toml` / `Cargo.lock` — workspace version bump.

## Tests

- **Differential** vs an independent recursive oracle across chain / diamond / multi-source / `None`-barrier shapes (guards behavior-equivalence).
- `kind == None` barrier, diamond union, and cycle-returns-`SchemaInferenceCycle`.
- **100k-node chain completes without stack overflow** — the core regression guard.
- Nested-subgraph descent + cycle termination for the incremental worklist.

## Test plan

- [ ] `cargo make clippy` clean
- [ ] `cargo test -p reearth-flow-runtime` green (incl. the new `dag_schemas::tests` and `incremental::tests`)
- [ ] CI engine workflow green

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -6162,7 +6162,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-log"
-version = "0.0.409"
+version = "0.0.410"
 dependencies = [
  "futures",
  "once_cell",
@@ -6182,7 +6182,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-plateau-processor"
-version = "0.0.409"
+version = "0.0.410"
 dependencies = [
  "approx",
  "async-trait",
@@ -6228,7 +6228,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-processor"
-version = "0.0.409"
+version = "0.0.410"
 dependencies = [
  "Inflector",
  "approx",
@@ -6286,7 +6286,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-python-processor"
-version = "0.0.409"
+version = "0.0.410"
 dependencies = [
  "indexmap 2.13.0",
  "once_cell",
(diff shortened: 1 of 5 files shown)


asrcpq (approved):
This PR does a forward propagation using the set of source node that affect some node as state (and set union as propagation operation), in a topologically sorted graph.

LGTM

PR #2178, opened 2026-06-18
fix(engine): normalize MVT winding instead of filtering

asrcpq:
# Overview

Because Flow does not require ring winding in its internal geometry representation, it is not correct to drop some specific winding but to normalize them based on exterior/interior property.

## Notes

This fix applies to MVT currently because it is a severe problem dropping valid geometry components.
However, many other formats currently do not do winding normalization, even they do not filter geometry, their output is considered potentially invalid. Including:

- 2D: MVT(this PR), GeoJSON
- 3D: obj, glTF, Cesium3DTiles, CityGML

## How I tested it

Using a problematic GeoJSON test and simply created a workflow to connect it to an MVT writer. After the fix, the previously invisible geometry is now rendered.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -6162,7 +6162,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-log"
-version = "0.0.412"
+version = "0.0.413"
 dependencies = [
  "futures",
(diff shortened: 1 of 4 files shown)


n4to4 (approved):
👍

PR #2189, opened 2026-06-19
fix(server): finalize preview-schema jobs from the sync probe response

pyshx:
## Summary

Fixes preview-schema jobs hanging forever at **`PENDING`** (the `jobStatus` subscription never resolves), the issue seen after #2180 fixed the report write.

## Root cause

A preview job is dispatched to the worker's `POST /probe-schema` route, which runs the **read-only** probe synchronously and returns the terminal status in the HTTP body — it does **not** publish a Redis `JobComplete` event like the `/run` path.

`PreviewSchemaCloudRunWorker` only acted on failure:
```go
status, err := i.cloudRunWorker.PreviewSchema(ctx, p)
if err != nil || status == gateway.JobStatusFailed { i.failCloudRunJob(...) }
// on COMPLETED: nothing — "trust the monitoring loop"
```
…but `checkJobStatus` finalizes a Cloud Run job only when a Redis worker event arrives (`statusChanged = batchStatusChanged || workerEvent != nil`; a preview job has an empty `GCPJobID`, so no Batch polling, and no event). So a **successful** probe never reached a terminal state — the job sat at `PENDING` until the 24h monitor timeout.

## Fix

Finalize preview jobs directly from the synchronous HTTP response (the authoritative result), instead of relying on the Redis/monitoring path:
- `status == COMPLETED` → new `completeCloudRunPreviewJob`: mark the job `COMPLETED`, run `handleJobCompletion` (surfaces the `SchemaReport` via `outputURLs`), and `subscriptions.Notify` so the `jobStatus` subscription resolves. Mirrors `failCloudRunJob` (no grace/Redis check — the probe is synchronous).
- failure → existing `failCloudRunJob`.

## Test

`TestJob_completeCloudRunPreviewJob` asserts a completed preview: job → `COMPLETED`, `completedAt` set, `outputURLs` = the report URL, **and** the subscription receives `COMPLETED` (directly guarding the forever-pending regression).

Build / `gofmt` / full `internal/usecase/interactor` suite green.

### DIFF
--- a/server/api/internal/usecase/interactor/job.go
+++ b/server/api/internal/usecase/interactor/job.go
@@ -199,8 +199,13 @@ func (i *Job) PreviewSchemaCloudRunWorker(j *job.Job, p gateway.ProbeSchemaParam
 			log.Errorfc(ctx, "job: preview-schema cloud run worker %s run error: %v", j.ID(), err)
 		}
 
-		if err != nil || status == gateway.JobStatusFailed {
+		// The probe is synchronous and emits no Redis event, so finalize the job
+		// directly from the HTTP response instead of relying on the monitoring loop.
+		switch {
+		case err != nil || status == gateway.JobStatusFailed:
 			i.failCloudRunJob(ctx, j.ID())
+		case status == gateway.JobStatusCompleted:
+			i.completeCloudRunPreviewJob(ctx, j.ID())
 		}
 	}()
 }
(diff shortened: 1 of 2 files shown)


pyshx:
Superseded by #2190. Rather than special-case completion in the server (finalize from the synchronous HTTP response), #2190 makes `probe-schema` publish a `JobCompleteEvent` so the existing monitoring loop finalizes the job — async, exactly like a debug run. No server change needed.

PR #2200, opened 2026-06-23
fix(ui): check for duplicate port names and refresh node [FLOW-DEV-209]

billcookie:
# Overview
This PR improves the UI editor’s router-port behavior by enforcing unique routing port names for router nodes and ensures React Flow nodes refresh their internal handle layout when the node’s ports (or collapsed state) change.
## What I've done
- Ensure `InputRouter` / `OutputRouter` nodes get a non-duplicated `routingPort` (based on existing pseudo-port names in the parent subworkflow node).
- Trigger `@xyflow/react` internal node refresh via `useUpdateNodeInternals` when input/output handle sets or collapsed state changes.


ui/src/lib/reactFlow/nodeTypes/GeneralNode/components/CustomHandle/Handles.tsx
```diff
@@ -44,6 +44,16 @@ const Handles: React.FC<Props> = ({
   const hasMoreThanFiveOutputHandles =
     outputs && outputs.length >= MIN_HANDLES_FOR_COLLAPSE;
 
+  const updateNodeInternals = useUpdateNodeInternals();
+  const handleSignature = useMemo(
+    () =>
+      `${inputs?.join(",") ?? ""}|${outputs?.join(",") ?? ""}|${nodeData.isCollapsed ? 1 : 0}`,
+    [inputs, outputs, nodeData.isCollapsed],
+  );
```
KaWaite, line 59: Why is this needed? Is it just to trigger the useEffect?

KaWaite (approved):

ui/src/lib/reactFlow/nodeTypes/GeneralNode/components/CustomHandle/Handles.tsx
billcookie: Yep so it runs when the handle set actually changes and not on every render and prevents the useEffect running unnecessarily. It will essentially only run if there is an addition, removal, renamed or if the node is collapsed.

If we want to guard further and prevent on first initial render could do something like this. React Flow already measures the handles on the first render (hence why it works when navigating between workflows or when you refresh):
```
const isFirstRender = useRef(true);
  useEffect(() => {
    if (isFirstRender.current) {
      isFirstRender.current = false;
      return;
    }
    updateNodeInternals(id);
  }, [id, handleSignature, updateNodeInternals]);
```

billcookie: Added a comment next to the code

PR #2215, opened 2026-07-01
fix(engine): fix _lod attribute for data-convert bldg workflow

asrcpq:
# Overview

This PR adapted #2050 to plateau5 branch to fix _lod attribute mismatch.

The actual problematic PR is #2006, merged earlier than #1953 (mentioned in #2050). `up to lod2 feature sent to lod4 branch` became _lod=4 but should be _lod=2.


n4to4 (approved):
👍

PR #2217, opened 2026-07-01
feat(engine): new CityGML2 reader tested with partially migrated PLATEAU4 workflows

asrcpq:
# Overview

This PR implements CityGML2 reader by reusing CityGML3 infrastructure. The previous FeatureCitygmlReader is PLATEAU-specific and will be removed in the future.

## Changes

- move citygml3 parser to a common `citygml-parser` module
- create `FeatureCitygml2Reader` as the next-generation alternative to `FeatureCitygmlReader` for reading Citygml 2.0 files
- Four new parser options in `FeatureCitygml{2,3}Reader`: `keepAttributes` (drop @-prefixed XML attributes), `flattenSingleChildObjects` (hoist single-child wrappers), `flattenMeasureTypes` (uom numeric elements = number + `{name}_uom` sibling), `cityGmlAttributesKey` (optional nesting key for parsed content; null = flat)
- migrate data-convert/{04,05,10,11} workflows to use `FeatureCitygml2Reader`
- fix luse/lsld test artifacts
- let PLATEAU4.attribute_flattener read `__citygml_feature_type` as fallback so `FeatureCitygml{2,3}Reader` do not need to manually duplicate another `featureType` attribute.

## Notes

FeatureCitygml{2,3}Reader's responsibility is not decided so these parameters may be moved to another action (or FlowExpr) implementation in the future. Currently they are implemented for enabling regression test.

Other unmigrated workflows mostly rely on new geometry, such as:

1. new lod representation and interface (used by 06-area-urf)
2. appearance parsing (used by 01,02,03, etc.)
3. lod4 subfeatures (used by 08-ubld)


engine/runtime/action-processor/src/citygml_parser/parser.rs
shunski: It looks like this it is stale, as it is now used by both 2.0 and 3.0.

shunski (commented):
It doesn't have to be addressed now, but
```
engine/runtime/action-processor/src/feature/reader/citygml2/processor.rs
engine/runtime/action-processor/src/feature/reader/citygml3/processor.rs
```
have a lot of duplicates.
Also, maybe we can make 
`engine/runtime/action-processor/src/feature/reader/citygml{2/3} `
a single file, rather than a folder module?

engine/runtime/action-processor/src/citygml_parser/parser.rs
asrcpq: remove tags - the new logging/error framework should provide which node/action triggers the error.

shunski (approved):
👍

PR #2234, opened 2026-07-08
chore(engine): enable schemars preserve_order for declaration-ordered action parameters

KaWaite:
## Overview

Enables the `preserve_order` feature on schemars so that generated action parameter schemas keep the Rust struct declaration order in their `properties` objects, instead of being alphabetized by schemars' internal BTreeMap.

## What changed

- `engine/Cargo.toml`: one-line feature addition (`preserve_order`) on schemars — this swaps its internal map to an insertion-ordered IndexMap
- All regenerated artifacts (`actions.json`, 5 language files, mdbook `action.md`): **key ordering changes only** — verified programmatically that content is identical to the previous generation for every action in all 6 schema files
- 62 actions now expose non-alphabetical (declaration) parameter order

## Why

This is a precursor to the action quality improvement PRs (see `engine/dev-docs/action-review-findings.md`). The Action Standard §3.5 requires parameters to be ordered by usability — required first, common optional next, edge-case optional last — which is impossible while the generator alphabetizes properties. Landing this separately keeps the mechanical reordering diff out of the content PRs.

## Verification

- Programmatic diff of old vs new JSON (order-insensitive) — identical content for all 155 actions × 6 files
- `cargo make clippy` — clean workspace-wide

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -2708,6 +2708,7 @@ checksum = "bd070e393353796e801d209ad339e89596eb4c8d430d18ede6a1cced8fafbd99"
 dependencies = [
  "autocfg",
  "hashbrown 0.12.3",
+ "serde",
 ]
 
 [[package]]
@@ -5792,6 +5793,7 @@ checksum = "3fbf2ae1b8bc8e02df939598064d22402220cd5bbcca1c76f7d6a310974d5615"
 dependencies = [
  "chrono",
  "dyn-clone",
+ "indexmap 1.9.3",
  "schemars_derive",
  "serde",
  "serde_json",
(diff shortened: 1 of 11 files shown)


KaWaite:
Re the two Copilot comments on `actions.json` (lines ~4370 and ~7552): both are false positives caused by diff-hunk misalignment.

`JsonWriter` has two structurally similar properties: `converter` (`type.enum: ["flowExpr"]`) and `output` (`type.enum: ["flowExpr", "string"]`). They swapped positions when property ordering changed from alphabetical to declaration order, so the line-based diff aligns old-`converter` against new-`output` — rendering as if the enum gained `"string"`. It didn't: both properties exist with unchanged enums in both versions, at both flagged locations.

Verified by parsing old and new JSON and comparing content order-insensitively — all 155 actions across all 6 schema files are semantically identical; the diff is key ordering only.

n4to4 (approved):
LGTM

PR #2240, opened 2026-07-09
refactor: rename base actions to space-separated title case

KaWaite:
## Overview

Renames all 74 base action IDs from PascalCase to **space-separated title case** per the Action Standard §1.1 — `CsvReader` → `CSV Reader`, `FeatureFilter` → `Feature Filter`, etc. The action `name` remains the single identity used everywhere (workflow YAML, UI display, docs), keeping one searchable name per action.

> **Stacked on #2236** (port rename) — review this PR's commits only; the base branch carries the port migration.

## Design decision

A separate `displayName` field was considered and rejected: dual naming means the UI shows one string while YAML/errors/docs/web answers show another, splitting searchability — and pre-release, ID renames are cheap (same one-time breakage window as #2236). If post-release rename freedom is ever needed, a display field can be introduced then, losing nothing.

## What changed

1. **`refactor(engine)`** — `fn name()` strings for the 74 base actions (+ router/filter action constants); hidden actions keep their names until reviewed
2. **`fix(worker)`** — user-facing log parser regexes captured action names with `(\w+)` (single word). With spaced names, the processor-error pattern stopped matching, so **a job whose processor errored would report "Workflow finished successfully" in the user-facing log**. Captures are now space-tolerant and still match single-word names identically
3. **`chore(engine)`** — fixture migration: all `action:` references in workflow YAML + golden logging fixtures (error *variant* identifiers like `FeatureCreator(...)` intentionally left as-is — they are Rust type names, not action names)
4. **`chore(engine)`** — regenerated schemas; verified programmatically: names remapped, zero content diffs, all 5 language translations followed their renamed keys
5. **`chore(server)`** — `base_actions.go` keys + tests
6. **`refactor(ui)`** — hardcoded `InputRouter`/`OutputRouter` literals; `encodeURIComponent` on all 5 action-URL construction sites (names now contain spaces; fetch auto-encodes, this makes it explicit)
7. **`docs`** — README examples; findings file headers renamed and all now-satisfied `name:` findings removed (every one matched the mechanical rename)

## Breaking change

Same class as #2236: saved projects/deployments referencing old action IDs fail at load and need re-creating — accepted pre-release, and landing in the same window as the port rename means users break once, not twice.

## Verification

- Engine: clippy, unit tests (incl. all 10 logging tests), 182/182 workflow integration tests
- UI: typecheck, 204/204 tests, lint, Prettier
- Server: `go test ./internal/app/`
- `test-conv` still requires the locally-missing `glb-decompress` binary — CI covers it


billcookie (approved):

n4to4 (approved):
LGTM 👍

PR #2247, opened 2026-07-13
feat(engine): add FeatureGeoJsonWriter processor and fix Point GeoJSON output

n4to4:
# Overview

Adds a `Feature GeoJSON Writer` processor and fixes CityGML `Point` geometries being silently dropped during GeoJSON conversion.

## What I've done

- Added the `FeatureGeoJsonWriter` processor.
- Fixed CityGML `Point` geometries being silently dropped during GeoJSON conversion.

## What I haven't done

- No `new-geometry` support yet (`finish` is gated off, like `FeatureWriter`).

## How I tested

- Added an integration test with a workflow fixture and input data.
- Ran `cargo make format`, `cargo make clippy`, `cargo make test`.

## Memo


shunski (approved):
👍

PR #2248, opened 2026-07-13
chore(engine): draco-oxide version update

shunski:
This PR updates the draco-oxide version.
The new version contains new API, performance refactor, and input filtering/repair.

# How it's tested
Visually verified that there is no regression in bldg/ubld conversion workflows with the Shinjuku dataset. 

# Changes to tiles test
The tiles test `examples/citygml-roundtrip/tun/cesium_statistics` failed on saying "mass-center not matching the truth" after the update. This is due to the fact that `draco-oxide` now automatically reduces the duplicate faces with identical orientation, as there is no visual difference by doing so. 
This is fixed by generating the new truth.


asrcpq (approved):
LGTM

PR #2259, opened 2026-07-15
feat(engine): add GeoJSON error geometry output to plateau6 bldg checks

n4to4:
# Overview

Adds GeoJSON error-geometry output to the plateau6 02-bldg quality checks:

- LOD0面の交差
- LOD2以上境界面のエラー

## What I've done

- `error_geometry/02_建築物_LOD0面の交差.geojson` — the intersection-area polygons computed by `AreaOnAreaOverlayer` are the error geometry. `AreaOnAreaOverlayer` now carries the input EPSG onto its outputs so the CRS is recorded.
- `error_geometry/02_建築物_LOD2以上境界面のエラー.geojson` — each feature is the geometry of the offending `con:*Surface` itself, so the error is pinpointed at surface level rather than per building. LOD2 and LOD3 merge into one file, mirroring the combined CSV.
- Both files carry the same identifying attributes as the CSV rows (`check`, `issue`, `gmlId`, `lod`, etc.), keep coordinates in the inspection CRS, and record it via a legacy `crs` member (`urn:ogc:def:crs:EPSG::{code}`).
- Supporting changes: `.geojson` verification in the workflow-tests harness, and a new `Z-bldg-06_nonplanar-wall-LOD2` test case demonstrating surface-level identification.

## What I haven't done

- Other checks are still CSV-only; GeoJSON output will be extended in follow-ups.

## How I tested

- `cargo test -p workflow-tests test_quality_check_plateau6_02_bldg` — all 36 tests passed.
- Added unit tests for the EPSG carry-over and the `crs` member generation — all passed.

## Memo


shunski (approved):
👍

PR #2263, opened 2026-07-16
feat(engine): 3D predicates for new geometry

shunski:
# Overview
This PR implements some 3D predicates for new geometry.

## Algorithms implemented in this PR
- predicate kernel for 3D: wraps `robust`'s exact predicates.
- 3D intersections of geometries: Exact intersection predicates between `Euclidean3DGeometry`, which is embedded as part of `intersects()`. For polygons and meshes, it is "exact" only in terms of its triangulation.
- Point position in 3D: Exact predicates for point-containment in solid.
- Ray casting: will be used by `Rayintersector3D`.
- Distance: the dimension-wise closest-pair distance for the the geometries implemented. 

## Limitations
The full 3D predicates and 3D binary operations need a highly complex library like CGAL (C++, paid license for commercial use). The algorithms implemented in this PR are only a tiny subset of it, which are usually enough for GIS computations.

## Scope
- CSG evaluation is intentionally out of scope for this PR, as it deserves its own PR.

## Notes
Legacy paths for the algorithms above aren't removed yet. They will be removed having migrated everything to new geometry.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4094,7 +4094,7 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "plateau-tiles-test"
-version = "0.2.76"
+version = "0.2.77"
 dependencies = [
  "bytes",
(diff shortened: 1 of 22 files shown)


asrcpq (approved):
LGTM

PR #2264, opened 2026-07-16
feat(engine): new cesium writer appearance, flip uv convention, and lod handling

asrcpq:
# Overview

This PR implemented basic appearance writing for the new cesium writer. It also handled lod parsing in the new CityGML as per-geometry attributes for split/filtering. Also, this PR flips the UV direction convention (new geometry only) to be top-left based.

## Why top-left?

- UV related implementations are simpler and more efficient. Atlas packer had to flip UV back and forth in the old convention.
- modern choice (Vulkan, WebGPU, DirectX, Metal vs OpenGL, WebGL)
- Natural choice to have (0,0) maps to `u=0`, `v=0`

drawback: during migration all texture/uv related reader/writer need to flip the order. But since we are migrating everything to new geometry this is probably the best chance to move to a better convention.

## Todo

- KTX2
- max atlas size + multipage export
- resolution knob (geometric error aware)
- PBR/Phong actual material parameter conversion (currently fixed parameter)
- theme selection in parameter
- double-sided material and multiple TEX_COORD attributes

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4632,7 +4632,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-log"
-version = "0.0.440"
+version = "0.0.441"
 dependencies = [
  "futures",
  "once_cell",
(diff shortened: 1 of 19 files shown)


shunski (approved):
👍

PR #2268, opened 2026-07-20
feat(websocket-go): make peer write queue size configurable, default 512

bnimit:
## What
Makes the ygo per-peer outbound broadcast queue depth (`PeerWriteQueueSize`) configurable and raises the default **256 → 512**.

- New env var **`REEARTH_FLOW_PEER_WRITE_QUEUE_SIZE`** (default 512), wired into both server setups (`ws.go`, `persistence.go`), following the existing `MaxPeersPerRoom` pattern.

## Why
When a peer's outbound queue fills, ygo **disconnects** it (forcing reconnect + full resync). We were on ygo's default of **256** and never overrode it, so under bursty updates we saw frequent `peer write queue full; closing slow peer` warnings and reconnect churn, even with just two peers in a room.

The Rust (yrs) server tolerates more before reacting and, notably, handles a lagging peer by **skipping** missed updates rather than disconnecting (shared `tokio::broadcast` buffer of **512**, `RecvError::Lagged` → continue). ygo only offers "disconnect on full," so the practical lever is a larger queue. Bumping the default to **512** matches the Rust buffer and gives legitimate clients more burst headroom; deployments can tune it further via the env var.

Not a data-loss issue (reconnect resyncs cleanly); this reduces churn and the associated log noise.

## Verification (`GOWORK=off`, in `server/websocket-go`)
- `go build ./...`, `go vet ./...` ✅
- `go test ./...` and `go test -race ./...` ✅ (all packages)
- Added config tests: default 512, env override, non-positive → default.

### DIFF
--- a/server/websocket-go/internal/config/config.go
+++ b/server/websocket-go/internal/config/config.go
@@ -42,6 +42,12 @@ type Config struct {
 	MaxPeersPerRoom int
 	// MaxRooms caps distinct rooms server-wide (doc_id is client-supplied).
 	MaxRooms int
+	// PeerWriteQueueSize is the per-peer outbound broadcast queue depth; when a
+	// peer's queue fills (slow/lagged connection) ygo disconnects it, forcing a
+	// reconnect-and-resync. Larger values tolerate bigger update bursts before
+	// evicting. ygo's own default is 256; we default higher to match the Rust
+	// server's 512 broadcast buffer.
+	PeerWriteQueueSize int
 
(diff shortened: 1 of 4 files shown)


bnimit:
Closing in favor of an ygo-side fix. This bumped our per-peer write-queue buffer to mask the churn, but the Rust (yrs) server handled slow peers fine at its defaults because it *lags and skips* rather than disconnecting. The right fix is graceful degradation in ygo's websocket provider (a lag-and-skip / overflow-policy option), not a larger buffer on the flow side. Tracking that on the ygo side instead.

PR #2275, opened 2026-07-21
feat(engine): implement fixed-size atlas and multi-page planner in atlas packer

asrcpq:
# Overview

This PR implements texel size and atlas size for cesium writer. It supports multi-page export and removes the adaptive downsampling trial loop. These are needed for ktx2 support (which will land with another PR).

## Changes

- `atlas/src/multipage.rs`: New multi-page atlas packer. Scales each texture to its own target resolution and spills overflow onto extra pages instead of shrinking the whole set. Errors on atlas sizes too small to hold one region.
- `atlas/src/lib.rs`: Gate the legacy single-page packer and the multi-page packer. Carry per-texture scale as an input field. Remap UVs per axis (frame/source ratio) so independently-rounded frame dimensions map exactly.
- `atlas/src/blit.rs`: Share the extrusion-fill helper with the new packer; gate the legacy blitter.
- `atlas/examples/pack_efficiency.rs`: Rewrite for the new multi-page packer.
- `cesium3dtiles/next/mod.rs`: Downsample textures by metres-per-pixel (texel size), pack into multi-page WebP atlases, split textured geometry one primitive per page.
- `cesium3dtiles/sink.rs`: Add `texel_size`, `atlas_size`, and `atlas_extrusion` writer parameters.
- `cesium3dtiles.rs`, `gltf.rs`, `action-sink/src/lib.rs`: Gate the legacy atlas-based writers (glTF, old Cesium pipeline).
- `examples/gml_to_3dtiles.rs`: Read texel size and atlas size from the environment.

## Screenshots

Textures (duplicated) resampled with texel size=0.01/0.1/1

### Planning strategy

When a placement fails, the current atlas is frozen and a new atlas will be used with next placement. Because for KTX2 the texture amount has less influence on texture size and packing efficiency

### Extrusion

Downsampling based method didn't have extrusion configurable because extrusion is there for solving bleeding caused by downsampling. The current method has pre-calculated scaling factor so does not have the same problem. The new configurable extrusion is used for mitigating bleeding during mipmap generation. Note that KTX2 pre-generates mipmap but still cannot solve mipmap bleeding is an empirical extrusion balancing atlas size and quality is still needed.

### Todos

- remove cropping and ktx2 support
- use jpg instead of webp as fallback. add format config
- add fixed extrusion config


shunski (approved):
👍

PR #2294, opened 2026-07-24
feat(engine): migrate Two Dimension Forcer to new geometry

n4to4:
# Overview

Migrates the `Two Dimension Forcer` processor to the new geometry stack (`new-geometry` feature), backed by a new `ForceTwoDimension` geometry op that drops the Z coordinate to produce pure 2D geometry.

## What I've done

- Added a `ForceTwoDimension` trait (`engine/runtime/geometry/src/ops/mod.rs`) that re-represents a 3D leaf as its matching 2D leaf (`Point3D` -> `Point2D`, etc.), keeping topology, appearance, and UV intact. Any 2.5D per-vertex elevation is cleared, so the op is idempotent.
- Implemented `force_2d` for `point`, `line_string`, `polygon`, `polygon_mesh`, and `triangular_mesh`; `Solid`, `Csg`, and `PointCloud` opt out via the trait default and report `UnsupportedOperation`.
- The coordinate frame (CRS tag) is preserved verbatim — no reprojection. This matches FME's `2DForcer`, which also only drops Z and offers no CRS option. Frame changes remain the responsibility of `Coordinate Frame Reprojector`.
- Wired the op into `enum_dispatch` and added `Geometry::force_2d` / `GeometryCollection::force_2d` (`engine/runtime/geometry/src/lib.rs`). Collection recursion is all-or-nothing: one member with no 2D counterpart fails the whole feature rather than silently dropping members.
- Added a `new-geometry` path to the processor that routes geometry with no 2D counterpart to a new `rejected` output port. The legacy path is unchanged.

## What I haven't done

- `PointCloud` opts out; a 2D bounding-box footprint (as FME does) is deferred.

## How I tested

- Added `force_2d` unit tests (`engine/runtime/geometry/src/lib.rs`) covering Z-drop, frame preservation, and appearance handling.
- Ran `cargo make schema-base` / `cargo make schema-translated`; no diff, so action schemas/i18n are in sync.

## Memo

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4212,7 +4212,7 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "plateau-tiles-test"
-version = "0.2.88"
+version = "0.2.89"
 dependencies = [
  "bytes",
  "gltf 1.4.1 (git+https://github.com/gltf-rs/gltf?rev=3def30c9caf96f2dbc85238179ea5a6152c5cf22)",
@@ -4750,7 +4750,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-log"
-version = "0.0.456"
+version = "0.0.457"
 dependencies = [
  "futures",
  "once_cell",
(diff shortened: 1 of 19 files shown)


shunski (approved):
👍

PR #2310, opened 2026-07-29
feat(engine): support Geometry Filter in new-geometry with detailed type mode

n4to4:
# Overview

Ports the `GeometryFilter` processor to the `new-geometry` world and adds a `detailedGeometryType` filter mode that routes features by their exact geometry type.

## What I've done

- Implemented `process` for `GeometryFilter` under `#[cfg(feature = "new-geometry")]`, routing on the new `Geometry` / `Euclidean2DGeometry` / `Euclidean3DGeometry` enums.
- Kept `none` and `geometryType` behaving as before, so existing workflows route to the same ports.
- Added `detailedGeometryType`, which gives each exact type its own port (`face`, `polygon-mesh`, `multi-surface`, `csg`, `aggregate`, …). The coarse mode collapses `Face` / `CompositeSurface` into one `surface` port; this mode separates them, matching what the FME reference workflows filter on.
- Dropped `multiple` / the `contains` port from the new-geometry build: it is not used by any existing workflow and has no FME counterpart. The legacy build is unchanged.

## What I haven't done

- No workflow-level integration test; coverage is unit tests only.

## How I tested

- Added a `new_geometry_tests` module covering every leaf type in both modes, collection unification (uniform / mixed / nested / empty), single- vs multi-member `GeometryCollection`, and the `polygon` vs `face` split.
- Added a test asserting every port a bucket can route to is declared by the factory — an undeclared port would silently drop features.
- `cargo make format` / `clippy` / `test` pass.

- The `Multi*` grouping in `multi_type_2d` / `multi_type_3d`: members are grouped by family, not exact type, so a GML multi-surface mixing polygons and oriented surfaces stays one `multi-surface`.
- Port naming: kebab-case (`multi-point`, `point-cloud`) vs the legacy camelCase names.

## Memo

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4212,7 +4212,7 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "plateau-tiles-test"
-version = "0.2.96"
+version = "0.2.97"
 dependencies = [
  "bytes",
  "gltf 1.4.1 (git+https://github.com/gltf-rs/gltf?rev=3def30c9caf96f2dbc85238179ea5a6152c5cf22)",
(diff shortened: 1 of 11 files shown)


asrcpq (approved):
LGTM

PR #2315, opened 2026-07-29
refactor(engine): apply action standard to Geometry B actions

KaWaite:
Applies [the action standard](engine/dev-docs/action-standard.md) to the **Geometry B (11)** group, the last content batch of the Phase 4 audit. Its section is deleted from `action-review-findings.md`, leaving only genuinely deferred items.

Behaviour was checked against the reference implementations these actions were ported from, against the PLATEAU workspaces that use them, and against OGC/PostGIS semantics where the operation is a standard geometry primitive.

> **Since #2343 made `new-geometry` a default feature, the generated schema now comes from that world.** Every metadata change here is live. Runtime behaviour is a different matter: five of these actions have no new-geometry `process` yet (Grid Divider, Polygon Normal Extractor, Ray Intersector, Refiner, Three Dimension Forcer) and two error by design (the reprojectors), so their behavioural fixes apply to the legacy path until each is ported. Called out per item below.

Also adds a **name-accuracy clause to §3.1** of the standard: a parameter name must describe what it actually controls, checked against the term the operation is known by outside this project. Two of the renames below come from applying it — a name can satisfy every existing §3.1 rule and still be wrong.

## Per-action changes

| Action | Description | Params | Ports | Tags |
|---|---|---|---|---|
| **Geometry Replacer** | rewritten | schema desc + `sourceAttribute` desc (documents that the attribute is consumed, and the rejected route) | **+`rejected`** (undecodable geometry), wired in **both** geometry worlds | `[]` → `attribute` |
| **Geometry Splitter** | rewritten | schema title/desc fixed on **both** param structs (each said "Parameters for GeometrySplitter"); `splitLevel` title + desc and `SplitLevel` variant titles on the legacy struct | **empty geometry now passes through** instead of being dropped (legacy path) | `split`,`geometry` → `[]` |
| **Geometry Validator** | rewritten | per-variant titles on `OptionalCheck`, `PlanarityThreshold` and `DegenerateThresholds` — **now the shipped parameter set**; `validationTypes` desc + `ValidationType` titles cover the legacy path | — | `validate` → `validation` |
| **Grid Divider** | rewritten | schema desc added (was missing), all 3 param descs sharpened; **`unitSquareSize`→`cellSize`, `keepSquareOnly`→`completeCellsOnly`** | — | `2d` → `spatial` |
| **Horizontal Reprojector** | rewritten | `sourceEpsgCode` trimmed from 4 sentences to 2; **reordered** required-first | — | `projection`,`2d` → `coordinate-system` |
| **Polygon Normal Extractor** | rewritten (now names the attributes it writes) | — | **+`rejected`** replaces job-killing errors | `normal`,`3d` → `3d` |
| **Ray Intersector** | period added | schema title/desc; **13 missing titles added**; per-variant titles on `OutputGeometryType`; **3 expression params → scalars, `pairId` → optional** | **`no_intersection` → `no-intersection`** | `ray`,`intersection`,`3d` → `3d`,`spatial` |
| **Refiner** | rewritten | — | **`remain` removed**; non-refinable geometry passes through | unchanged (`3d`) |
| **Three Dimension Forcer** | rewritten | schema title de-internalised, both param descs tightened; **`preserveExistingZ` default false → true** | — | unchanged (`3d`) |
| **Two Dimension Forcer** | rewritten | — | — | `2d` → `[]` |
| **Vertical Reprojector** | rewritten | schema desc; per-variant title + desc on `VerticalReprojectorType`, with a `TODO` for further datums per §3.4 | — | `projection`,`3d` → `coordinate-system`,`3d` |

## Behaviour and interface changes

**Three Dimension Forcer defaulted to destroying Z.** `preserveExistingZ` defaulted to **false**, so unless you knew to set it, the action overwrote Z on geometry that already had it. Adding a dimension should never discard coordinates already present — PostGIS `ST_Force3D` is explicit that *"Nothing happens to an already 3D geometry."* Now defaults to **true**. No workflow in this repo uses the action.

**Refiner's `remain` port duplicated the input stream.** `process` sent *every* feature to `remain` unconditionally and *then* emitted refined parts on `features`. The reference implementation has a single output port, and the corresponding node in the PLATEAU workspace is wired with exactly one output. Port removed; geometry that cannot be refined now passes through on `features` rather than disappearing. `REMAIN_PORT` is deleted from `node.rs` — Refiner was its only consumer engine-wide.

**Geometry Splitter silently discarded features.** `if geometry.is_empty() { return Ok(()); }` dropped features with no port at all, violating §4.3. The action's own `new-geometry` branch already passes non-splittable geometry through on `features`, so the two worlds disagreed; legacy now matches it.

**Polygon Normal Extractor no longer kills the job.** It previously returned `Err` on 2D/None/non-polygon input and on CityGML with >1 polygon, aborting the entire run over one stray feature. Those now route to `rejected` — matching every Extractor the new-geometry migration has touched. This action is not yet ported, so the routing takes effect on the legacy path; the `rejected` port is declared now so whoever ports it has it to wire.

**Geometry Replacer gains `rejected`, in both geometry worlds.** An undecodable stored geometry previously either passed out on `features` as though it had been transformed (legacy) or failed the whole run (new-geometry, after #2308). It is now routed:

| Source attribute | Outcome |
|---|---|
| absent | pass through on `features` — nothing to restore |
| present, not a string | pass through on `features` — nothing to restore |
| present, will not decode | `rejected` |

Worth a reviewer's eye: **#2308 migrated this action while the branch was open**, codifying pass-through for absent *and* non-string with tests. Since `get_output_ports` is not `cfg`-gated, a legacy-only `rejected` port would have been declared but never emitted under new-geometry — the same §4.3 defect this PR removes from Refiner. Both paths are reconciled instead. `replace()` keeps its signature and all five of its tests; only its caller changed. The absent case must stay a pass-through: `Geometry Extractor` skips features with empty geometry, so its counterpart legitimately receives features with nothing to restore.

**Ray Intersector asked for expressions where nobody had one.** `closestIntersectionOnly`, `includeRayOrigin` and `tolerance` were `FlowExpr`-typed but constant in every real use, so expressing "true" cost a three-line block. They are now plain `bool`/`bool`/`f64`. `pairId` was **required**, yet every caller passed a constant meaning "don't pair" — it is now optional, and omitting it tests every ray against every geometry. A node's `with` block goes from 13 lines to 4:

```yaml
# before                          # after
pairId:                           closestIntersectionOnly: true
  type: flowExpr                  includeRayOrigin: false
  value: '"all"'                  tolerance: 0.01
closestIntersectionOnly:
  type: flowExpr
  value: 'true'
# …and two more blocks like it
```

`geomId` stays an expression because it genuinely is one (`attributes.get("gmlId","")`). The contrast that makes this safe: Horizontal Reprojector's EPSG codes really are expressions — `env["prcs"]` in most uses — and are left alone.

**Grid Divider's two names described the wrong thing.** A "unit square" has side 1, so `unitSquareSize: 5.0` contradicts itself; the established term is cell (or edge) size. And every cell is a square — the distinction `keepSquareOnly` was drawing is complete vs partial. Now `cellSize` and `completeCellsOnly`.

## Breaking changes

Per the alpha no-migrator policy:

1. `Ray Intersector` port `no_intersection` → `no-intersection` (§4.1). 2 fixture files swept.
2. `Ray Intersector` — `closestIntersectionOnly`, `includeRayOrigin`, `tolerance` retyped from expression to scalar; `pairId` required → optional. 4 fixture nodes.
3. `Grid Divider` — `unitSquareSize` → `cellSize`, `keepSquareOnly` → `completeCellsOnly`. 4 fixture nodes.
4. `Three Dimension Forcer` — `preserveExistingZ` default flips false → true. No fixture uses it.
5. `Refiner` port `remain` **removed**. No fixture used it.
6. `Horizontal Reprojector` parameter order swapped (schema-only; serde is order-insensitive).

## Checks

| Check | Result |
|---|---|
| `cargo make format` | pass |
| `cargo make clippy` | pass, 0 warnings |
| `cargo make test-rs` (both worlds) | **3175 passed**, 0 failures |
| `cargo make test-qc` | 4 passed / 186 ignored — main's current state, since #2331 `#[ignore]`s every case still marked `skipNewGeometry` |
| `cargo test -p workflow-tests -- --test-threads=4` (legacy world) | **184 passed / 0 failed / 6 ignored** |
| `cargo make schema-base` re-run | no-op — canonical i18n order intact, so `check-schema` passes |
| `cargo make test-conv` | not run — needs `glb-decompress`; CI covers |

`test-qc` no longer exercises the legacy path, so the legacy workflow suite was run separately — most of this batch's behavioural fixes land there for now. Note it needs `--test-threads=4`: at full thread count 13 cases fail on a shared XML-schema-cache race, unrelated to this branch (they pass individually, with and without these changes, and the failures are all `L02エラー` / schema-import errors). Worth someone's attention as a separate issue now that nothing runs this suite by default.

**i18n** — Japanese translated by hand for all 11 actions; English placeholders in es/fr/zh refreshed; genuine es/zh translations left alone. Verified the diff touches only these 11 actions per language file.

---

# Deferred, with reasons

All of the below is recorded in `engine/dev-docs/action-review-findings.md` under `## Geometry B — deferred items only`, so nothing is lost if none of it is settled before this merges. Items 1–4 want a decision from a maintainer; 5–7 are recorded conclusions.

**1. Removing the two legacy reprojectors.** `Horizontal Reprojector` and `Vertical Reprojector` each already carry a `#[cfg(feature = "new-geometry")] process` that errors with *"use Coordinate Frame Reprojector instead"*, above a `TODO(new-geometry)` to delete the action. They were **audited rather than deferred** because the replacement is gated out of the shipped build — both the module (`geometry.rs:13`) and its factory registration (`geometry/mapping.rs:117`) sit behind the feature flag. **This is now actionable and arguably urgent.** With `new-geometry` defaulted on, `Coordinate Frame Reprojector` is in `actions.json` and can go into `base_actions.go` — while the two legacy reprojectors are in the base set, visible to users, and error on every feature. Task: delete both files, swap them for `Coordinate Frame Reprojector` in `base_actions.go`, migrate 29 + 20 fixture files. Out of scope here (server-side change plus a large fixture migration), but it should not sit long.

**2. `Coordinate Frame Reprojector` needs its own audit when ungated.** Two issues already visible: `epsgCode` is required only when `destinationFrame` is `crs`, which the schema cannot express as declared (`Option<u16>`, validated in `build`) — folding it into a tagged enum on `destinationFrame`, as `BasePoint` already is in the same file, would make it structurally required. And `basePointSource` is titled "Base Point" while the nested `BasePoint::Value.base_point` is *also* titled "Base Point", so the UI shows the label twice. Separately, FLOW-DEV-182 says *"Reprojector will be added instead"* but the action is named `Coordinate Frame Reprojector` — worth settling which name ships.

**3. Geometry Validator has two opposed parameter models, and four dead parameters today.** The dead-parameter half of this **resolved itself with the flag flip**: the four new params are now the live, shipped set and `validationTypes` has dropped out of the schema. The deeper issue remains — the two models are opposites: `validationTypes` is opt-**in** with tolerances inline, the new set is opt-**out** with separate thresholds. The opt-out model is right — it matches OGC, where validity is a property of the geometry and a separate detail call reports which rule failed. Converging the legacy path costs something real: **26 Validator nodes here each run exactly one check**, so `failed` currently means "failed *that* check". Under the new model it means "failed anything", and each node needs a downstream filter on `validationResult.checks` — worth confirming FlowExpr can even address a nested attribute like that.

**4. Polygon Normal Extractor — status and output design.** FLOW-DEV-182 has it **On hold** because *"PlanarityFilter implicitly does the same thing."* Only partly true: `PlanarityFilter` writes `surfaceNormalX/Y/Z` + `pointOnSurfaceX/Y/Z` and has no equivalent for `slope`, `azimuth` (conventioned `0° = South` to match `SunPositionCalculator`) or `signedArea2D`; it is also a filter, not an extractor. Audited on that basis — say the word if you would rather it reverted to a deferral. Separately, it is the only Extractor with no output-naming parameter: six fixed attribute names, and multi-polygon results smeared across numbered suffixes rather than one feature per polygon. That is a redesign, entangled with the On-hold question, so it belongs in one follow-up.

**5. Refiner should fold into Geometry Splitter** — their descriptions are not distinguishable by a user, and no workflow uses Refiner. Not done here because it is not a fold: legacy Splitter passes `GeometryCollection` straight through, which is the only case Refiner implements, so merging means *adding* collection support to the legacy path — new capability the new-geometry world discards anyway, since its generic `Split` already covers collections, meshes and point clouds. It also reaches into `base_actions.go`. Belongs at the flip.

**6. UI fix, shared with Geometry A.** `consolidateOneOfToEnum` in `patchSchemaTypes.ts` bails when a `oneOf` mixes string and object variants, so RJSF labels variants "option 1/2/…". Affects **Geometry Validator's `ValidationType`** here and **Image Rasterizer's `OnOverlap`** in #2317. Rust-side `/// # Title` fixes are done in both; one UI PR covers both actions, filed once.

**7. Considered and rejected.** Merging Two Dimension / Three Dimension Forcer behind a mode enum: 2D Forcer has a `rejected` port in the migrated world and ports cannot vary by parameter, so a merged action would declare a port that sits dead half the time. Renaming the Forcers ("Three Dimensions Forcer" / "Three-Dimensional Forcer" per the findings): the established names for this operation elsewhere are single tokens, so neither the current names nor either alternative collides, and the rename buys nothing — **decided: keep both names as they are**, closing the only `name:` finding in the section.


asrcpq (approved):
LGTM

PR #2317, opened 2026-07-29
refactor(engine): apply action standard to Geometry A actions

KaWaite:
Applies [action-standard.md](engine/dev-docs/action-standard.md) to the 12 **Geometry A** actions. Companion to #2315 (Geometry B) — these are the last two content batches of the Phase 4 audit.

Behaviour was checked against the reference implementations these actions were ported from, and against OGC/JTS/PostGIS semantics where the operation is a standard geometry primitive.

## Per-action changes

| Action | Changed |
|---|---|
| Appearance Remover | desc |
| Area Calculator | desc; 3 missing param titles; `AreaType` per-variant docs (the old text named the variants in PascalCase, but the values are camelCase); tags `[area, measurement]` → none; **always writes its attribute, 0 where there is no area** |
| Bounds Extractor | desc; schema-level desc; param descs; **accepts an absent `with` block** |
| Bufferer | desc; tags `[2d]` → `[spatial]`; `bufferType` TODO; `interpolationAngle` desc corrected; **no longer panics on CityGML** |
| Clipper | desc; tags `[2d]` → `[spatial]` |
| Elevation Extractor | desc; param desc; **stops writing a fabricated 0 for geometry with no elevation** |
| Extruder | desc; param desc; **`rejected` port replaces job-killing `Err`** |
| Footprint Replacer | desc (dropped the implementation parenthetical) |
| Geometry Extractor | desc; param desc noting the Geometry Replacer pairing |
| Geometry Part Extractor | desc; `geometryPartType` kept with a TODO + variant title; port doc comments; tags → `[3d]` |
| Geometry Remover | desc |
| Image Rasterizer | desc; tags → `[raster]`; `imageWidth` title/desc; **all 5 `OnOverlap` variants get titles**; **ports → kebab-case**; **`rejected` port**; **accepts an absent `with` block** |

## Breaking changes

- Image Rasterizer ports `textureCoordinates` → `texture-coordinates` and `textureBounds` → `texture-bounds` (§4.1). Three fixture edges updated.

## Two new `rejected` ports

Extruder and Image Rasterizer. Neither can lose data that flows today — Extruder previously returned `Err` and failed the whole job, and Image Rasterizer dropped the features silently — but they may need routing in workflows that care.

Two other actions were considered for a `rejected` port and deliberately did **not** get one. Geometry Extractor and Elevation Extractor treat "nothing to extract" as a no-op and pass the feature through, because a new port is unwired in every existing workflow and would silently delete those features. Geometry Extractor's pass-through is load-bearing: its counterpart Geometry Replacer relies on receiving features that carry no stored geometry.

## On `geometryPartType` (single-variant enum, kept)

The parameter has one option, `surface`, so §3.4's design-smell rule applies. It is **kept with a TODO** rather than removed, on the same footing as Bufferer's `bufferType`.

`edge` and `vertex` variants — emitting each edge or vertex as its own feature — fit the three ports this action already declares, and nothing else covers them: Boundary Extractor returns the boundary as one geometry on the same feature, and Coordinate Extractor writes vertices into attributes. Removing the parameter would have foreclosed that space and pushed us toward a separate action per part type, against the goal of a concise action set.

Related constraint worth recording, since it will come up again: `get_output_ports` takes only `&self`, so **ports cannot vary by parameter value**. Actions can only be consolidated behind a mode enum when they share a port shape — otherwise the merged action must statically declare the union of ports and most sit dead, which §4.3 treats as a bug.

## Findings that turned out to be stale

- **Bounds Extractor param ordering.** The finding says the schema is alphabetical; it is already axis-grouped `xmin, xmax, ymin, ymax, zmin, zmax`. `preserve_order` (#2234) fixed it. No change.
- **Geometry Part Extractor's `remaining` vs `untouched`.** The finding asks whether they are distinct. They are — `remaining` is the original feature with its extracted parts removed, `untouched` is a feature the action did not modify. Documented, not changed.

## Deferred, with reasons (recorded in the findings file)

- **Bufferer only buffers points, curves and single polygons.** Everything else — multi-polygons above all — is emitted unbuffered. This deviates from the standard model: JTS defines `buffer()` on the base `Geometry` type and *"the buffer operation always returns a polygonal result"*, so no type is un-bufferable. **Consequence: a distance tolerance is silently not applied to those features, including the 0.005 near-touching tolerance in the PLATEAU `surface_validator` graph.** Not fixed here because it moves quality-check results and the truth data needs someone who can adjudicate PLATEAU conformance. Worth its own issue. (The projection to 2D on that path is correct and stays — PostGIS: *"This function ignores the Z dimension. It always gives a 2D result even when used on a 3D geometry."*)
- **Bufferer's `bufferType` keeps its single-variant `oneOf`** — a solid buffer type is genuinely missing, but needs a solid-buffering algorithm the geometry crate does not have.
- **Image Rasterizer's `features` output means two different things** (pass-through of `texture-coordinates` input, or a synthetic feature holding the image path). Entangled with the known `HOME`-fallback save-path issue; both belong in one follow-up.
- **The `OnOverlap` variants still render as "option 1/2/…" in the UI** even now that every variant has a title, because `consolidateOneOfToEnum` in `patchSchemaTypes.ts` bails when a `oneOf` variant is an object type. UI-side fix, separate PR.
- **`CityGmlAttributeInserter`'s input port `textureBounds`** is camelCase too, but it is a PLATEAU action outside the base set. Its edge now joins two differently-named ports until it is reviewed.

## Out-of-batch audit note

`boundary_extractor.rs` gains a comment only — no code change. While reading it to settle the `geometryPartType` question, its default path looked like it drops features whose boundary cannot be extracted, with no port receiving them and CityGML worst affected. Boundary Extractor is hidden and unaudited, so the note is written as leads for its future auditor to **verify**, not as conclusions to apply.

## Notes

- Descriptions deliberately avoid naming the current geometry variants, so they stay accurate once the unified geometry model lands. Elevation Extractor's change converges on it: the new model's planar leaves return `elevation() == None`, which is why writing 0 was wrong.
- Behavioural fixes land in the `cfg(not(new-geometry))` path only for the eight actions not yet migrated; whoever migrates them needs to carry them over. The four already migrated (Appearance Remover, Bounds Extractor, Geometry Extractor, Geometry Remover) agree across both worlds.
- Merged `main` at 0.0.467; the two Remover descriptions were resolved in favour of the wording added on main, with the house Oxford comma applied.

## Checks

`cargo make format` · `cargo make clippy` (both lines, incl. `--features new-geometry`) · `cargo make test-rs` (both lines) · `cargo make test-qc` — **184 passed, 0 failed**. `test-conv` needs `glb-decompress`, not available locally; CI covers it.


n4to4 (approved):
👍

PR #2320, opened 2026-07-30
feat(engine): 3D rasterization test for PLATEAU6 01-bldg data-convert workflow

asrcpq:
# Overview

This PR implements basic 3D rasterization test for PLATEAU6 01-bldg workflow and configures CI test for new geometry plateau-tiles-test (needed by PLATEAU6 workflows).

## Algorithm

- Camera config schema added to `profile.toml`: `[convs.raster3d.<id>.cameras.<name>]` with ECEF `position`, `look_at`, `up`, `fov_y_deg`, `near`, `far`.
- `render3d` module: `Camera3d` + view-proj construction, `rasterizer::render_depth` rasterizes triangles into a depth buffer.
- `input::recenter_and_cast` recenters triangle coordinates to the camera position before casting `f64`→`f32`, avoiding precision loss at ECEF scale.
- Fixes a Cesium-specific axis relabeling: glTF content embedded in 3D Tiles stores ECEF `(x,y,z)` as `(x,z,-y)`; `load_triangles` flips it back to true ECEF.
- Depth buffer is stored as inverse depth (`1/distance`, background `0.0`) rather than raw distance, so absolute error matters less the farther the surface is.
- Canvas is serialized via `write_png_f32`/`read_png_f32`, packing each pixel's raw `f32` bits into RGBA8 channels — lossless but not human-viewable, unlike the 8-bit grayscale PNGs used by the 2D raster tests.
- `Canvas::compare_depth` scoring: per pixel, take the `(min,max)` range over its 3×3 neighborhood (out-of-bounds = background `0.0`) in both canvases, score `max(a_min-b_max, b_min-a_max, 0)`, aggregate via RMS.

## Screenshots

Ground truth images are f32 depth images encoded in PNG RGBA8 (32bit grayscale is possible with tiff/exr but need external dependency). 

cam1

## Notes

- Since current implicit tiling implementation is add-only, loading all triangles is the correct method (also used for statistical test). If that changes, the loader strategy (also statistical test) also needs change.


n4to4 (approved):
👍

PR #2321, opened 2026-07-30
feat(engine): support Hole Counter and Hole Extractor in new-geometry

n4to4:
# Overview

Make `Hole Counter` and `Hole Extractor` work under the `new-geometry` feature, backed by a new pair of geometry operations (`CountHoles` / `ExtractHoles`) that share one definition of "hole".

### Hole Counter

- Writes the number of interior rings to `outputAttribute`; every feature leaves via `features`.
- Counting is total: geometry that cannot carry a hole — point, curve, point cloud, triangle mesh, absent geometry — yields `0`.
- A `Solid` counts the holes in the faces of every shell. A void shell is a hollow volume, not a face boundary, so it is not itself a hole.

### Hole Extractor

- Each face is taken apart: outer boundary with holes removed to `outershell`, each interior ring to `hole` as an area of its own. A face without holes still leaves via `outershell`, unchanged.
- Meshes, solids, and collections deaggregate — one feature in, many out — each part keeping the input's attributes and taking a fresh id.
- Rings are emitted verbatim, neither re-wound nor re-closed: a hole's winding is exactly what the surface validation workflows inspect.
- Geometry bounding no area goes to `rejected`, member-by-member inside a container.

### Differences from FME

Matches FME on the behavior that matters (zero for non-area geometry; outershell passthrough, deaggregation, non-area rejection). Not replicated: FME emits a hole containing a smaller hole via both ports (nested donuts) — here each ring is emitted once. FME's `Aggregate Handling` parameter is not exposed; deaggregation is fixed, as in the existing action.

### Differences from the legacy implementation

| | legacy | new-geometry |
| --- | --- | --- |
| Counter, no geometry | forwarded without the attribute | attribute set to `0` |
| Counter, triangle mesh | panics (`unimplemented!()` in `runtime/geometry/src/algorithm/hole.rs`) | `0` |
| Extractor, input types | polygon / multi-polygon / CityGML only, else rejected whole | any area geometry, meshes and solids included |
| Extractor, mixed container | one non-area member rejected the whole feature | only that member is rejected |

## What I haven't done

- No new end-to-end workflow test; covered by unit tests at the geometry-op and processor level.
- No schema / i18n regeneration — parameters and ports are unchanged.

## How I tested

- Geometry-op unit tests in `runtime/geometry/src/ops/hole.rs`, with counting and extraction driven from shared fixtures plus a cross-check that the reported count equals the number of extracted `hole` parts.
- Processor unit tests asserting the exact port-and-feature sequence, attribute preservation, and fresh part ids.
- `cargo make format`, `cargo make clippy`, `cargo make test`, with and without `new-geometry`.

- Impact on the consumers: `runtime/examples/fixture/graphs/surface_validator.yml` and the 10 `quality-check/plateau4/*.yml` workflows, where the count feeds a `Feature Filter` on `attributes["holeCount"] > 0` and the extractor feeds the interior orientation / intersection checks.
- Whether a `Solid`'s void shells should be counted as holes (currently not — only the rings inside their faces).

## Memo

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4212,7 +4212,7 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "plateau-tiles-test"
-version = "0.2.97"
+version = "0.2.98"
 dependencies = [
  "bytes",
  "glam 0.30.10",
@@ -4751,7 +4751,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-log"
-version = "0.0.468"
+version = "0.0.469"
 dependencies = [
  "futures",
  "once_cell",
@@ -4771,7 +4771,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-plateau-processor"
-version = "0.0.468"
+version = "0.0.469"
 dependencies = [
  "approx",
  "async-trait",
@@ -4815,7 +4815,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-processor"
-version = "0.0.468"
+version = "0.0.469"
 dependencies = [
  "Inflector",
  "approx",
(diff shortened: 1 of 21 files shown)


asrcpq (approved):
LGTM

PR #2330, opened 2026-08-03
feat(engine): cesium writer targetTileSize and sub-tile splitting

asrcpq:
- [ ] FIXME: update schema after new-geometry schema generation merge 

# Overview

This PR replaces cesium writer's max depth parameter with tile size. Max depth instead becomes a builtin constant as a safety cap. Implements sub-tile splitting to incrementally load big tiles from several small fragments. Also adjust 01-bldg to split geometry by grouping lod to fix the oom bug.

## Changes

- `splitter.rs`: add an optional `groupBy` parameter grouping geometry collection members into the same output feature.
- `cesium3dtiles/next/`: replace the `maxDepth` parameter with a target tile size; coarse per-feature byte cost estimates drive merging small cells and splitting oversized ones into same-tile contents.
- `cesium3dtiles/next/mod.rs`: quadtree placement depth is now a hardcoded safety limit of 24, no longer user-tunable.
- `workflow.yml`: group LOD geometries by `lod`, switch 3D Tiles writer nodes to target tile size.

## Demo

[tileset_quadtree.html](https://github.com/user-attachments/files/30695367/tileset_quadtree.html)

Note that this over-estimated bytes-per-texel. This is fixed in later update (1B->0.2B)

## Known limitations

- currently tile size estimation (for triggering split) is coarse. bytes estimation in fundamentally hard unless you rebuild a tile multiple tiles wasting CPU. The better way is to run time estimate the multiplier to minimize retry attempts, which is advanced and belongs to another PR.

- not a limitation but splitter flattens one level (nested geometry collection goes up 1 level). This means group_by key is valid and still preserving the structure.

- group_by is partially reimplementing the aggregator (but avoiding attributes explosion). A general solution is to have a "reaggregator" that generalize both splitter and aggregator with group key list (also with geometry collection idx as a valid key) so splitter=(feature_idx, geometry_idx) as key.

### Other issues

This PR fixes OOM issue on tokyo CG3 example dataset.
The direct cause is Arc<Attributes> spilled to tmpfs. But it leads to some (not urgent) concerns:

1. disk spilling is /tmp on Linux. This will cause problem once docker container goes up to Ubuntu 24.04 where /tmp becomes tmpfs. Fix should be easy but need to decide a better path.
2. disk spiling/feature intermediate data write cannot respect Arc<Attributes> share - could silently explodes disk while RAM is small.

## Notes

Intra-cell split is not implemented yet. Also both splits still require atomic feature so cannot solve the problem that features cross the boundary gets unconditionally loaded:

download

An option is to add a toggle to use centroid instead of bbox to lower the placing to leaf cell (which can cause problem when looking a big feature from close distance, the feature maybe not loaded).


engine/runtime/action-sink/src/file/cesium3dtiles/sink.rs
```diff
@@ -179,12 +179,13 @@ pub struct Cesium3DTilesWriterParam {
...
-    /// follows feature size. Defaults to 24, from 0 to 24.
+    /// # Target Tile Size
+    /// Target content size per tile, in bytes. Tiles are split when they'd
+    /// exceed it and merged with neighbours when they'd otherwise be smaller;
+    /// a single feature that alone exceeds it is kept whole (features are
+    /// never clipped). Defaults to 1,048,576 (1 MiB).
     #[cfg(feature = "new-geometry")]
-    #[schemars(range(min = 0, max = 24))]
-    pub(super) max_depth: Option<u32>,
+    pub(super) target_tile_size: Option<u64>,
```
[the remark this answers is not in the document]
asrcpq: this is desired actually. so the actual fix becomes: doc this behavior.

engine/runtime/action-processor/src/geometry/splitter.rs
```diff
@@ -266,6 +301,88 @@ impl Processor for GeometrySplitter {
...
+            let single = group_members.len() == 1;
+            // A lone member is the unambiguous owner of its attrs, so they promote
+            // to the feature. A merged group has no single owner among its
+            // members' attrs, so the feature-level attrs are left untouched; each
+            // member's own attrs stay attached on the nested `GeometryCollection`.
+            let representative_attrs = if single {
+                group_attrs.first().cloned().unwrap_or_default()
+            } else {
+                Default::default()
+            };
```
shunski: For any group with two or more members, an empty `Attributes` is merged in. Thus, if the collection have multiple elements, it would lose the group key. It is `lod` in the plateau 6 case, all features are lost when the LoD group has multiple members.

shunski: Confirmed the fix

shunski (approved):
👍

Issue #2337, opened 2026-08-03
Document/websocket GraphQL operations have no workspace authorization check


bnimit:
Every operation on the document/websocket surface takes a `projectId` and passes it straight to the websocket HTTP client with no workspace or project permission check. Authentication is enforced (the GraphQL endpoint sits behind `authMiddleware`, `internal/app/app.go:115-117`), but authorization is not: **any authenticated user can act on any project in any workspace.**

### Root cause

`interfaces.Container.Websocket` is typed `WebsocketClient` (`internal/usecase/interfaces/common.go:38`) and is wired to the bare infrastructure client:

```go
// internal/usecase/interactor/common.go:60
Websocket: client,
```

Every other Container entry is a permission-checking interactor. Compare `internal/usecase/interactor/project.go:56`, which calls `checkPermission(ctx, rbac.ResourceProject, ...)`. The websocket entry bypasses that layer entirely, so no resolver on this surface has anywhere to inherit a check from.

### Affected operations

All in `internal/adapter/gql/resolver_document.go`:

| Operation | Impact |
|---|---|
| `projectHistory` | read another workspace's version metadata |
| `latestProjectSnapshot` | read another workspace's document state |
| `projectSnapshots` | read snapshot labels, timestamps, sizes |
| `previewSnapshot` | read arbitrary historical document state |
| `saveNamedSnapshot` | write into any project |
| `rollbackProject` | **destructive** — reaches `PruneAfter`, deleting every update above the target clock |

`rollbackProject` is the most serious: an unauthorized caller can permanently truncate another workspace's project history.

Two concrete abuse paths beyond plain reads:

- **Cross-tenant information disclosure.** Snapshot labels are user-authored and often descriptive ("pre-launch config", customer or project names), so listing them leaks business information even without fetching state.
- **Destroying a victim's named versions.** Snapshot retention is bounded per room (`KeepSnapshots`, default 50), and ygo's trim keeps the newest N regardless of label. Repeatedly calling `saveNamedSnapshot` against a victim project evicts their deliberately named snapshots. See reearth/ygo#212 for the label-aware-retention side of this.

### Scope note

[...]

Issue #2336, opened 2026-08-03
Version panel read-only guard test cannot detect the regression it claims to guard


bnimit:
The test asserting that snapshot rows in the Version panel stay read-only passes against a component that reintroduces a row click handler, so it provides no protection.

### Why the assertions cannot fail

`ui/src/features/Editor/components/OverlayUI/components/Version/components/VersionHistoryList.test.tsx:418-432` carries the comment "If a click handler is reintroduced here, this must start failing", then asserts:

1. `queryAllByRole("button")` has length 0
2. clicking a row does not throw

A `<div onClick={...}>` has no implicit `button` role, and clicking it does not throw. Both assertions hold with a click handler present, so neither can detect the named regression.

Confirmed by mutation: adding `cursor-pointer` and `onClick={() => alert(snapshot.id)}` to the row `div` (`VersionHistoryList.tsx:56-58`) leaves the file reporting **4 passed / 0 failed**.

### Why this matters more than a normal weak test

The read-only constraint is not cosmetic. `NamedSnapshot.id` (a per-room snapshot counter) and the update-log `version` (a CRDT clock) are unrelated integer spaces, and the revert path reaches `PruneAfter`, which deletes every update above the target clock. Feeding a snapshot id into it means permanent data loss. A regression of exactly this kind was caught during review of the snapshot work, which is why the rows were made read-only.

So this test will be cited as evidence the invariant holds. A guard that cannot fail is worse than no guard, because it removes the incentive to check.

### Fix

Assert the contract rather than the rendering:

- Spy on `useDocument().useGetPreviewProjectSnapshot` and the rollback mutation, `fireEvent.click` every row, and assert **0 calls**. That directly encodes "no snapshot id can reach preview or revert".
- Optionally also assert the row has no `tabindex` and no `cursor-pointer` class.

Validate the replacement by re-running the mutation above: it must fail.

### Related

The structural constraint itself is currently correct — no path exists from a `NamedSnapshot.id` to preview or revert, verified by grep across `ui/src`. This issue is only about the test that is supposed to keep it that way. Snapshot-addressed preview/restore is tracked in #2332.

bnimit:
Folded into #2335, which now covers the whole UI follow-up as a single unit of work: the recovery-dialog crash, this guard test, the "auto"-label problem, the error-vs-empty state, missing i18n, and the panel minors.

[...]

Issue #2335, opened 2026-08-03
UI follow-up: Version panel and recovery dialog after snapshot-backed history


bnimit:
Follow-up UI work for the snapshot-backed version history landing in #2338 (backend). Grouped as one issue because it is one PR: everything here is in the Version panel or the recovery dialog beside it.

Two of these are pre-existing bugs on `main`, not regressions from the snapshot work — noted per item.

---

## 1. The recovery dialog throws on open (pre-existing on `main`)

`RecoveryDialog.tsx:47` calls `useEditorContext()`, which throws when there is no provider:

```ts
// ui/src/features/Editor/editorContext.tsx:47-53
const ctx = useContext(EditorContext);
if (!ctx) throw new Error("Could not find EditorProvider");
```

The dialog renders from the route's `errorComponent` (`routes/workspaces.$workspaceId_.projects_.$projectId.lazy.tsx:39-41`, `:167-174`), which **replaces the entire route subtree**, so it renders outside `<Editor>`. Every `EditorProvider` in the app is a *descendant* of this dialog, never an ancestor. The context is guaranteed absent on exactly the path that renders it.

Effect: clicking "Revert to a previous version" on the corruption screen throws while rendering the error component, so the user gets a blank or root-level error instead of the dialog. Since this is the only revert path in the product, a user whose project will not open has no way back to a working version.

Confirmed pre-existing: at the merge base, `errorComponent` already rendered `VersionDialog` and `Version/index.tsx:28` already made the same call. The snapshot branch renamed and split the component, carrying the bug along unchanged.

`isLocked` is meaningless here anyway — a project that will not open cannot be usefully edit-locked.

**Fix:** drop `useEditorContext()` and the `isLocked` term from the Revert button's `disabled`, or read defensively via `useContext(EditorContext)` with `isLocked ?? false`. Add a regression test rendering `ProjectRecoveryDialog` with **no** `EditorProvider` ancestor — the production configuration, currently uncovered, which is why this survived.

## 2. The read-only guard test cannot fail

`VersionHistoryList.test.tsx:418-432` says "If a click handler is reintroduced here, this must start failing", then asserts `queryAllByRole("button")` is empty and that clicking does not throw. A `<div onClick>` has no implicit `button` role and clicking it does not throw, so neither assertion detects the named regression.

Confirmed by mutation: adding `cursor-pointer` and `onClick={() => alert(snapshot.id)}` to the row div leaves the file at **4 passed / 0 failed**.

[...]

Issue #2334, opened 2026-08-03
Remove dead interactor.Websocket (TODO 2026-09)


bnimit:
Tracking the `TODO(2026-09)` left in `server/api/internal/usecase/interactor/websocket.go`.

`interactor.Websocket` is unreachable in full, not in part:

- `NewWebsocket` has no callers anywhere in the tree
- `interfaces.Container.Websocket` is typed `WebsocketClient` (`server/api/internal/usecase/interfaces/common.go:38`), so the GraphQL resolvers hold the client, not this wrapper

The clearest tell is that `resolver_document.go` calls `CopyDocument` and `ImportDocument`, which do not exist on `interactor.Websocket` at all — it exposes `CopyProject` / `ImportProject`, and both are unreachable. It compiles only because the container field never had this type.

It was left in place for one release rather than deleted immediately, in case something out-of-tree constructs it. The tree itself does not.

### To do

- Confirm nothing outside this repo depends on `interactor.NewWebsocket`
- Delete `server/api/internal/usecase/interactor/websocket.go`
- Check whether `interfaces.WebsocketClient` still needs every method it declares; `GetHistory` in particular has no caller either

No behaviour change is expected, since nothing executes this code today.

PR #2338, opened 2026-08-03
feat(api, websocket): snapshot-backed version history (backend)

bnimit:
## Why

The Version history panel has been showing only the current version. Root cause traced end to end, and it is not a missing feature — it is a coupling problem:

The panel derived its rows from the raw CRDT **update log**, one row per persisted update object. ygo v1.36.0 (#2283) then enabled **write coalescing** by default: a 2s debounce that resets on every update, capped at 10s, merging each burst into a single `StoreUpdate`. `LegacyAdapter.StoreUpdate` is one `AppendUpdate` is one GCS object is one history row, so the row count went from *per committed transaction* to *per idle-gap*. A light session (open a project, place two nodes, close) now produces one or two objects, and the UI renders the newest as "Current Version" and filters it out. Result: an empty list.

Both user complaints are one bug with the sign flipped. Under yrs, one write per transaction made every drag tick its own version row — that was the "versions switching every second" churn. Coalescing fixed the write volume and the history went with it.

So the fix is not to write more often. It is to stop deriving user-visible versions from a write-batching log. This PR stores real snapshots, so version count is independent of persistence tuning.

Ruled out along the way, to save a reviewer re-deriving it: `ListSnapshots` has no checkpoint filter (only `filterByCeiling`), `compactToCheckpoint` (`clock%10==0`) writes doc_v2 and the checkpoint but deletes nothing, and `projectHistory` is unlimited end to end. Restoring yrs-style update pruning (`KeepVersions`/`CompactEvery`) was considered and rejected: it acts on the same axis coalescing already reduced, so it would shorten the list further.

## What

Backend only. Implements ygo v1.41.0's `SnapshotStore` and `RoomLister` on the GCS adapter, enables server-side auto-versioning, and exposes snapshots over HTTP and GraphQL.

- **Storage** — snapshot objects with label and uncompressed size in custom metadata, so `ListSnapshots` is one `listAttrs` call that never reads a payload. State is stored as an **opaque brotli blob**, never decoded as a CRDT update: `SnapshotStore`'s contract and its conformance suite allow any non-empty byte string. Keeping it V1 also means a Rust rollback can still read these blobs.
- **Auto-versioning** — `AutoVersionEvery` (default 15m) and `KeepSnapshots` (default 50), both env-overridable. ygo versions a room at most once per interval and only when it changed, and force-versions on room unload when dirty. So any session that changed something yields at least one real version at close, a quiet room is never versioned, and history accumulates per work session.
- **API** — `GET/POST /api/document/{id}/snapshots`, `GET /api/document/{id}/snapshots/{sid}`, a `projectSnapshots` query and a save mutation.

## Review found four things worth calling out

A four-reviewer pass over the whole branch found two silent data-loss paths in the storage layer. Both are fixed here with tests verified to fail against a mutant restoring the old behaviour.

**A lost id counter overwrote existing snapshots** (live on the default path). The counter was the sole source of truth and the write was a plain put, so a room whose counter went missing restarted allocation at id 1 and overwrote live snapshot 1 — payload, label and timestamp gone, no error anywhere. Routinely reachable: `Delete` removed the counter *before* sweeping the snapshots it numbers, and those deletes are sequential and unlocked, so a cancelled request stranded snapshots with no counter. With auto-versioning on, the history was then eaten 15 minutes at a time. Fixed in three layers: the counter is a fast path only and falls back to deriving the floor from stored objects; snapshot records are written **create-only** with a `DoesNotExist` precondition so a colliding id fails loudly instead of destroying bytes; and `Delete` removes the counter last.

**Flipping `REEARTH_FLOW_GCS_PHASE2` hid every existing room's history.** Snapshot reads used the primary layout only, while every other layout-scoped read here already falls back to the legacy root. Document state kept loading fine via dual-read, so the flag flip looked clean while the history panel silently emptied. Reads now merge both layouts, and the allocator seeds from `max(primary, legacy)`. Not currently live — the flag defaults off and is not set in the infrastructure repo — but it was a trap waiting for whoever flipped it.

**The endpoints reported phantom successes.** An empty document returned `200 {"id":0}` for a snapshot that does not exist (ids start at 1), reachable by clicking Save version on an unedited project; a malformed body silently dropped the user's label; unsupported backends degraded three different ways including a 404 that claimed a snapshot was missing rather than the feature absent; and the state response reused a shape whose `version` field means an update-log clock, shipping `"version":0` into a payload where feeding 0 to rollback prunes the entire log.

**The kill switch failed toward enabled.** `REEARTH_FLOW_AUTO_VERSION_EVERY` fell back to the 15m default on anything unparseable, so `-1`, `off` and `15` all left auto-versioning running while looking like they disabled it. Now a hard startup error, following the existing `WS_PROTECTED` precedent.

## Testing

`GOWORK=off go test ./...` green in `websocket-go`, `golangci-lint` 0 issues, gofmt clean. `server/api` builds and `./internal/...` is green. The `e2e` package needs MongoDB and was not run locally. UI typechecks unchanged.

Every fix above is pinned by a test verified to fail against a mutant. One nuance recorded in the code: the end-to-end counter tests pass even against a counter-trusting mutant, because the precondition and retry mask the bad allocation, so `nextSnapID` has its own direct assertion.

## Not in this PR

The UI half is deliberately separate — it has known open items and I would rather not hold the storage fixes behind them. Reviewers here need the GCS layout and crash-safety code; reviewers there need the panel.

Filed from this review pass:

- #2337 — **document/websocket operations have no workspace authorization check.** Pre-existing across the family, and `rollbackProject` is the worst of it. This PR adds two operations to the same pattern rather than creating it. Worth prioritising independently of this work.
- #2332 — snapshot-addressed preview/restore. Rows are read-only for now because `NamedSnapshot.id` and the update-log `version` are unrelated integer spaces and revert reaches `PruneAfter`.
- #2333 — admin cleanup enumerates only live rooms, so retention never reaches the idle rooms it exists for. `gcs.Adapter.ListRooms` is implemented and conformance-tested but deliberately unwired pending that decision.
- #2334 — delete the unreachable `interactor.Websocket`.
- #2335, #2336 — UI: the recovery dialog throws on open (pre-existing on `main`), and the read-only guard test passes against a click handler.
- reearth/ygo#212 — `trimSnapshots` keeps the newest N regardless of label, so a user's deliberately named snapshot is evicted once N newer auto ones exist. Roughly half a day of editing at 15m/50.
- reearth/ygo#211 — the snapshot conformance suite never exercises room names needing encoding, which is the class of bug that bit the id parsing here.


server/websocket-go/internal/config/config.go
```diff
@@ -142,6 +172,17 @@ func (c *Config) Validate() error {
...
+	// is ON by default, so this var is the kill switch an operator reaches for
+	// during an incident. envDuration falls back to the 15m default on anything
+	// it cannot parse, which means "-1", "off", "disabled" and "15" (no unit) all
+	// leave the feature running while looking like they turned it off. Use "0"
+	// to disable.
+	if raw := os.Getenv("REEARTH_FLOW_AUTO_VERSION_EVERY"); strings.TrimSpace(raw) != "" {
+		if _, err := time.ParseDuration(strings.TrimSpace(raw)); err != nil {
+			return fmt.Errorf("REEARTH_FLOW_AUTO_VERSION_EVERY=%q is not a valid Go duration (e.g. 15m, 30s; use 0 to disable auto-versioning); refusing to start rather than silently keeping auto-versioning enabled", raw)
+		}
+	}
```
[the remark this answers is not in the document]
bnimit: Taken, with one correction to the reasoning.

**The panic risk does not apply here.** There is no ticker built from this value. ygo reads it once at worker construction and disables the feature outright:

```go
// ygo provider/websocket/persistence.go
autoVersionEvery := s.AutoVersionEvery
if autoVersionEvery <= 0 {
    versioner = nil
}
```

The interval is then compared inline (`clock.now().Sub(lastVersion) < autoVersionEvery`), so `-1m` already behaved correctly — it disabled auto-versioning.

**But the suggestion is still right, for a different reason.** It worked silently, and only by coincidence. `0` is the documented way to disable, so anything else non-positive is far likelier a stray minus or a copied `-1` than an intent — and the entire reason this var is validated rather than allowed to fall back is that an operator reaching for a kill switch during an incident must never have their intent guessed at. Accepting `-1m` quietly is the same fail-open class the check exists to prevent, just with a benign outcome.

Fixed in f5417767c: negative durations are now rejected, with `0` kept as the only permitted non-positive value.

Two test gaps this exposed, also fixed: the auto-versioning defaults and env overrides were pinned by nothing, so a typo in either key would have silently reverted the feature to its default while looking configured; and `clearEnv` did not unset the two new vars, so ambient environment leaked into every config test.

server/api/internal/adapter/gql/resolver_document.go
```diff
@@ -50,6 +50,25 @@ func (r *queryResolver) ProjectHistory(ctx context.Context, projectId gqlmodel.I
...
+		return nil, err
+	}
+
+	nodes := make([]*gqlmodel.NamedSnapshot, len(snaps))
+	for i, s := range snaps {
+		nodes[i] = &gqlmodel.NamedSnapshot{
+			ID:        int(s.ID),
+			Label:     s.Label,
+			Timestamp: s.Timestamp,
+			Size:      int(s.Size),
```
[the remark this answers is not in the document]
bnimit: Partly taken — `size` was genuinely wrong, `id` I am pushing back on.

**`size`: fixed, and the real defect was inconsistency.** This schema already has a `FileSize` scalar mapped to `graphql.Int64`, and `Asset.size` uses it. So there were two byte counts in one API typed two different ways for no reason. `NamedSnapshot.size` is now `FileSize!` and the resolver no longer narrows — the int64 flows straight through.

Worth noting the overflow itself was not reachable: `size` is an uncompressed CRDT document, and a 2 GiB one would break far more than this field (it is held in memory and broadcast over the websocket). The narrowing *would* have truncated on a 32-bit build, and matching the existing scalar costs nothing, so it is fixed on consistency grounds rather than risk.

**`id`: staying `Int!`, deliberately.** It is a per-room snapshot counter starting at 1. Exhausting 32 bits needs ~2.1 billion snapshots in a single room; at the 15-minute auto-version interval that is on the order of tens of thousands of years. There is no bound check worth adding either — erroring on a valid snapshot because its id is large would introduce a failure mode where none exists. Widening it would imply a risk that is not there.

**The real hazard on that field is not its width.** `NamedSnapshot.id` and `ProjectSnapshotMetadata.version` are both `Int!` in the same schema file, and they are unrelated ID spaces: `version` is a CRDT update-log clock, and `rollbackProject(projectId:, version:)` reaches `PruneAfter`, which deletes every update above the target clock. Passing a snapshot id there prunes at an unrelated point and destroys history — a regression of exactly that kind was caught during review of this work, which is why the UI rows are read-only.

So I have documented that on the field instead, since it is the thing a future caller can actually get wrong. Making the two mutually unassignable at the type level (a distinct scalar, or `ID`) is the stronger fix and I would rather do it when clients start consuming snapshot ids — tracked in #2332, where preview and restore get wired up.

Fixed in f5417767c.

server/api/internal/adapter/gql/resolver_document.go
```diff
@@ -108,6 +127,20 @@ func (r *mutationResolver) ImportProject(ctx context.Context, projectId gqlmodel
...
+	s, err := usecases(ctx).Websocket.SaveNamedSnapshot(ctx, string(projectId), label)
+	if err != nil {
+		return nil, err
+	}
+
+	return &gqlmodel.NamedSnapshot{
+		ID:        int(s.ID),
+		Label:     s.Label,
+		Timestamp: s.Timestamp,
+		Size:      int(s.Size),
```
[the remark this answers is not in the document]
bnimit: Partly taken — `size` was genuinely wrong, `id` I am pushing back on.

**`size`: fixed, and the real defect was inconsistency.** This schema already has a `FileSize` scalar mapped to `graphql.Int64`, and `Asset.size` uses it. So there were two byte counts in one API typed two different ways for no reason. `NamedSnapshot.size` is now `FileSize!` and the resolver no longer narrows — the int64 flows straight through.

Worth noting the overflow itself was not reachable: `size` is an uncompressed CRDT document, and a 2 GiB one would break far more than this field (it is held in memory and broadcast over the websocket). The narrowing *would* have truncated on a 32-bit build, and matching the existing scalar costs nothing, so it is fixed on consistency grounds rather than risk.

**`id`: staying `Int!`, deliberately.** It is a per-room snapshot counter starting at 1. Exhausting 32 bits needs ~2.1 billion snapshots in a single room; at the 15-minute auto-version interval that is on the order of tens of thousands of years. There is no bound check worth adding either — erroring on a valid snapshot because its id is large would introduce a failure mode where none exists. Widening it would imply a risk that is not there.

**The real hazard on that field is not its width.** `NamedSnapshot.id` and `ProjectSnapshotMetadata.version` are both `Int!` in the same schema file, and they are unrelated ID spaces: `version` is a CRDT update-log clock, and `rollbackProject(projectId:, version:)` reaches `PruneAfter`, which deletes every update above the target clock. Passing a snapshot id there prunes at an unrelated point and destroys history — a regression of exactly that kind was caught during review of this work, which is why the UI rows are read-only.

So I have documented that on the field instead, since it is the thing a future caller can actually get wrong. Making the two mutually unassignable at the type level (a distinct scalar, or `ID`) is the stronger fix and I would rather do it when clients start consuming snapshot ids — tracked in #2332, where preview and restore get wired up.

Fixed in f5417767c.

bnimit:
## Merge-order note: interacts with #2341

#2341 fixes the missing workspace authorization on this same GraphQL surface (issue #2337), and the two PRs collide on `server/api/internal/usecase/interactor/websocket.go`.

This PR extends `interfaces.WebsocketClient` with `GetSnapshots` and `SaveNamedSnapshot`. #2341 turns `interactor.Websocket` from dead code into an authorizing wrapper that implements that interface. So whichever merges second **needs a follow-up commit to compile**.

### Suggested order: #2341 first

It is smaller, self-contained, and a security fix. This PR then rebases and adds the two new methods to the wrapper, `ActionRead` for `GetSnapshots` and `ActionEdit` for `SaveNamedSnapshot`, matching how their siblings are treated.

### Worth being explicit about the risk

Both operations added here (`projectSnapshots`, `saveNamedSnapshot`) are currently **unauthorized**, like every other operation on this surface. If this PR merges first and the conflict is later resolved by leaving them on the raw client, they stay unauthorized while everything around them is guarded, which is the worst outcome: it looks fixed and is not. `websocket_permission_test.go` in #2341 is table-driven precisely so a missing row fails rather than passes quietly.

Also note this PR marks `interactor.Websocket` as dead code with a `TODO(2026-09)` pointing at #2334. #2341 supersedes that by giving the type a real purpose, so **#2334 can be closed** rather than actioned.

server/api/gql/document.graphql
```diff
@@ -28,6 +28,21 @@ type ProjectSnapshotMetadata {
   version: Int!
 }
 
+# A labelled, named snapshot in a project's version history (distinct from
+# ProjectSnapshot, which is a raw update-vector snapshot keyed by version).
+type NamedSnapshot {
+  # A backend-assigned, per-room snapshot counter. NOT interchangeable with
+  # ProjectSnapshotMetadata.version, which is a CRDT update-log clock: passing
+  # this id to rollbackProject or previewSnapshot would prune at an unrelated
+  # clock and destroy history. See reearth-flow#2332.
```
pyshx: Please remove these comments, the types itself are self-explanatory.

server/api/gql/document.graphql
```diff
@@ -28,6 +28,21 @@ type ProjectSnapshotMetadata {
...
+  id: Int!
+  label: String!
+  timestamp: DateTime!
+  # FileSize (int64), matching Asset.size — the uncompressed document byte count
+  # comes off the wire as an int64 and GraphQL Int is only 32-bit by spec.
```
pyshx: remove

server/api/gql/document.graphql
```diff
@@ -28,6 +28,21 @@ type ProjectSnapshotMetadata {
   version: Int!
 }
 
+# A labelled, named snapshot in a project's version history (distinct from
+# ProjectSnapshot, which is a raw update-vector snapshot keyed by version).
```
pyshx: Remove

server/api/internal/adapter/gql/resolver_document.go
```diff
@@ -50,6 +50,28 @@ func (r *queryResolver) ProjectHistory(ctx context.Context, projectId gqlmodel.I
...
+	if err != nil {
+		return nil, err
+	}
+
+	nodes := make([]*gqlmodel.NamedSnapshot, len(snaps))
+	for i, s := range snaps {
+		nodes[i] = &gqlmodel.NamedSnapshot{
+			// Size stays int64 (FileSize scalar). ID narrows to int, which is safe:
+			// it is a per-room snapshot counter, so exhausting 32 bits would take
+			// billions of snapshots in a single room.
```
pyshx: remove

server/api/internal/infrastructure/websocket/client.go
```diff
@@ -458,6 +458,127 @@ func (c *Client) DeleteDocument(ctx context.Context, docID string) error {
...
+	}
+
+	snapshots := make([]*websocket.SnapshotMetadata, len(snapshotsResp))
+	for i, item := range snapshotsResp {
+		// Deliberately NOT falling back to time.Now(): the panel sorts rows by
+		// timestamp and uses it as the row label when a snapshot is unlabelled, so
+		// a fabricated "now" would sort an old snapshot above genuinely newer ones
+		// and mislabel it. The websocket server always writes RFC3339, so a parse
+		// failure is a protocol mismatch worth surfacing — leave the zero value,
+		// which sorts last rather than first.
```
pyshx: remove

server/api/internal/infrastructure/websocket/client.go
```diff
@@ -458,6 +458,127 @@ func (c *Client) DeleteDocument(ctx context.Context, docID string) error {
...
+
+	return snapshots, nil
+}
+
+// SaveNamedSnapshot captures the document's current state as a new labelled
+// snapshot. The underlying store's SaveSnapshot only returns (int64, error),
+// so the save endpoint's response carries just {id, label}; Timestamp and
+// Size are the zero value there. We enrich the result from GetSnapshots here
+// so callers (and ultimately the GraphQL client) get complete metadata
+// without needing a second round trip of their own.
```
pyshx: Single liner is enough

server/api/internal/infrastructure/websocket/client.go
```diff
@@ -458,6 +458,127 @@ func (c *Client) DeleteDocument(ctx context.Context, docID string) error {
...
+	if err := json.NewDecoder(resp.Body).Decode(&saved); err != nil {
+		return nil, fmt.Errorf("failed to decode response: %w", err)
+	}
+
+	// Snapshot ids are allocated from 1, so a non-positive id means the server
+	// created nothing however cheerful the status code was. Returning it as a
+	// success would hand the UI a version row addressing no stored snapshot,
+	// which then disappears on the next refetch. Defensive on purpose: the
+	// websocket server now 409s this case, but the API server must not depend on
+	// a specific server version to avoid inventing history.
```
pyshx: same as above

server/api/internal/infrastructure/websocket/client_snapshots_test.go
```diff
@@ -0,0 +1,156 @@
...
+	assert.False(t, got[1].Timestamp.IsZero())
+}
+
+// An unparseable timestamp must leave the zero value rather than substituting
+// time.Now(). The Version panel sorts rows by timestamp and, for an unlabelled
+// snapshot, renders the timestamp AS the row label — so a fabricated "now" would
+// float a stale snapshot above genuinely newer ones and label it with today's
+// date. The zero value sorts last and reads as obviously wrong, which is the
+// honest failure mode. The row itself is still returned: one bad timestamp must
+// not drop a snapshot the user can otherwise see and restore.
```
pyshx: same as above

server/api/internal/infrastructure/websocket/client_snapshots_test.go
```diff
@@ -0,0 +1,156 @@
...
+	assert.NoError(t, err)
+
+	_, err = client.GetSnapshots(context.Background(), "proj1")
+	assert.Error(t, err)
+}
+
+// TestClient_SaveNamedSnapshot_EnrichesFromList exercises the enrichment path:
+// the save endpoint only returns {id, label} (Timestamp/Size zero-valued,
+// mirroring the underlying store's SaveSnapshot signature), so the client
+// must call the snapshot list to fill in Timestamp and Size before returning.
```
pyshx: same as above

server/api/internal/infrastructure/websocket/dto.go
```diff
@@ -35,3 +35,17 @@ type snapshotResponse struct {
 type importDocumentRequest struct {
 	Data []int `json:"data"`
 }
+
+// snapshotItemResponse is the wire shape for one labelled snapshot, returned
+// both by GET .../snapshots (fully populated) and POST .../snapshots (only ID
+// and Label populated; Timestamp/Size are the zero value there).
```
pyshx: same as above

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -7,6 +7,14 @@ import (
...
 )
 
+// TODO(2026-09): delete this file — tracked in reearth/reearth-flow#2334.
+// Websocket is dead code — NewWebsocket has no
+// callers, and interfaces.Container.Websocket is typed WebsocketClient, so the
+// GraphQL resolvers talk to the client directly (note they call CopyDocument /
+// ImportDocument, which do not even exist on this type; the CopyProject /
+// ImportProject wrappers below are unreachable). Kept for one release in case
+// something out-of-tree constructs it; remove once that is ruled out. Do not add
+// new methods here — add them to interfaces.WebsocketClient instead.
```
pyshx: A single liner TODO is enough

server/api/pkg/websocket/websocket.go
```diff
@@ -19,3 +19,11 @@ type HistoryMetadata struct {
 	Timestamp time.Time
 	Version   int
 }
+
+// SnapshotMetadata is one labelled snapshot in a project's version history.
```
pyshx: Not needed.

server/websocket-go/internal/config/config.go
```diff
@@ -51,6 +51,30 @@ type Config struct {
 	// queue overflow instead of disconnecting. Default ON; env REEARTH_FLOW_SLOW_PEER_RESYNC.
 	SlowPeerResync bool
 
+	// AutoVersionEvery, when > 0, asks ygo to capture a labelled snapshot of a
+	// room at most this often and only when the room changed, giving the Project
+	// History panel meaningful entries instead of one per CRDT update.
+	// 0 disables auto-versioning.
```
pyshx: please use a single liner comment.

server/websocket-go/internal/config/config.go
```diff
@@ -51,6 +51,30 @@ type Config struct {
...
+	// Two limits worth knowing, both from ygo's LegacyAdapter.trimSnapshots:
+	//
+	//  1. It runs only on the AUTO path (SaveVersion). Snapshots created through
+	//     POST /api/document/{id}/snapshots call the store directly and are not
+	//     trimmed, so with auto-versioning disabled the manual route can grow a
+	//     room's snapshots without bound (there is no DELETE endpoint either).
+	//  2. The trim keeps the newest N for the room REGARDLESS of label, so a
+	//     user's deliberately named snapshot is evicted once N newer ones exist.
+	//     At 15m/50 that is roughly half a day of continuous editing. Label-aware
+	//     retention needs an upstream change: reearth/ygo#212.
```
pyshx: Single Liner is enough.

server/websocket-go/internal/config/config.go
```diff
@@ -142,6 +172,26 @@ func (c *Config) Validate() error {
 			return fmt.Errorf("REEARTH_FLOW_WS_PROTECTED=%q is not a valid boolean (use true/false, on/off, yes/no); refusing to start with an ambiguous WS auth setting", raw)
 		}
 	}
+	// Same fail-loud reasoning as above, for a different reason: auto-versioning
+	// is ON by default, so this var is the kill switch an operator reaches for
+	// during an incident. envDuration falls back to the 15m default on anything
+	// it cannot parse, which means "-1", "off", "disabled" and "15" (no unit) all
+	// leave the feature running while looking like they turned it off. Use "0"
+	// to disable.
```
pyshx: Too much verbose on this comment, Please put one liner if necessary

server/websocket-go/internal/config/config.go
```diff
@@ -142,6 +172,26 @@ func (c *Config) Validate() error {
...
+	if raw := os.Getenv("REEARTH_FLOW_AUTO_VERSION_EVERY"); strings.TrimSpace(raw) != "" {
+		d, err := time.ParseDuration(strings.TrimSpace(raw))
+		if err != nil {
+			return fmt.Errorf("REEARTH_FLOW_AUTO_VERSION_EVERY=%q is not a valid Go duration (e.g. 15m, 30s; use 0 to disable auto-versioning); refusing to start rather than silently keeping auto-versioning enabled", raw)
+		}
+		// A negative duration parses fine and ygo treats <= 0 as "disabled", so it
+		// would work — but silently, and only by coincidence. "0" is the documented
+		// way to disable, so anything else non-positive is more likely a typo (a
+		// stray minus, a copied "-1") than an intent, and this var's whole purpose
+		// is that an operator's intent is never guessed at.
```
pyshx: same as above

server/websocket-go/internal/config/config_test.go
```diff
@@ -334,6 +335,66 @@ func TestValidateAcceptsValidAndEmpty(t *testing.T) {
 	}
 }
 
+// TestAutoVersioningDefaultsAndOverrides pins the two auto-versioning knobs.
+// Neither was covered, so nothing held the 15m/50 defaults or proved the env
+// names parse at all — a typo in either key would have silently reverted the
+// feature to its default while looking configured.
```
pyshx: same as above

server/websocket-go/internal/config/config_test.go
```diff
@@ -334,6 +335,66 @@ func TestValidateAcceptsValidAndEmpty(t *testing.T) {
...
+	}
+}
+
+// TestValidateRejectsBadAutoVersionEvery: auto-versioning is ON by default, so
+// this var is the kill switch an operator reaches for during an incident.
+// envDuration falls back to the 15m default on anything it cannot parse, which
+// would leave the feature RUNNING while looking disabled — the same fail-open
+// trap Validate already guards for WS_PROTECTED. A negative duration is rejected
+// too: ygo happens to treat <= 0 as disabled, so it would work, but only by
+// coincidence and silently, and "-1" is far likelier a typo than an intent.
```
pyshx: same as above

server/websocket-go/internal/gcs/kv.go
```diff
@@ -40,6 +42,49 @@ func (s kv) put(ctx context.Context, name string, data []byte) error {
 	return w.Close()
 }
 
+// putWithMeta writes data and attaches custom object metadata, so listAttrs can
+// return a snapshot's label and uncompressed size without reading its payload.
```
pyshx, line 46: Single liner.

server/websocket-go/internal/gcs/kv.go
```diff
@@ -40,6 +42,49 @@ func (s kv) put(ctx context.Context, name string, data []byte) error {
...
+}
+
+// errObjectExists reports that a create-only write lost to an existing object.
+var errObjectExists = errors.New("gcs: object already exists")
+
+// putWithMetaIfAbsent is putWithMeta that REFUSES to overwrite, returning
+// errObjectExists instead. Snapshot records are write-once by contract (ids are
+// never reused within a room), and a plain put makes an id collision destroy the
+// previous snapshot's payload, label and timestamp with no error anywhere. The
+// precondition converts that silent data loss into a caller-visible failure.
```
pyshx: same as above.

server/websocket-go/internal/gcs/layout.go
```diff
@@ -108,6 +113,53 @@ func (LegacyRootLayout) UpdatePrefix(_ DocID, oid uint32) string {
...
+// SnapVersionIDFromName recovers the snapshot id from a hex-encoded object name,
+// and reports false for anything that is not a snapshot record.
+//
+// The marker check is load-bearing, not defensive dressing. Without it, taking
+// "everything after the last colon" makes any all-decimal tail parse as an id —
+// so a room whose name hex-encodes to digits has its own COUNTER object read as
+// a snapshot (room "0" yielded (30, true), room "12345" yielded (3132333435,
+// true)). Today's only caller pre-filters by SnapVersionPrefix, but that makes
+// correctness depend on caller discipline for a method on the Layout interface,
+// and a caller listing a broader prefix would silently mint phantom ids.
```
pyshx: same as above.

server/websocket-go/internal/gcs/rooms.go
```diff
@@ -0,0 +1,97 @@
...
+// useless for retention since the rooms needing cleanup are the idle ones. That
+// swap is a behaviour change and is tracked in reearth/reearth-flow#2333 rather
+// than widened into the snapshot work. Kept and conformance-tested so the swap
+// is a one-line change when it happens.
+//
+// Phase 2 is a straight prefix listing. Phase 1 has no folder structure: object
+// names are hex-encoded structured keys, so rooms are recovered from the OID
+// index objects (one per room, name = hex(V1‖KEYSPACE_OID‖utf8(room)‖0x00)) plus
+// the snapshot-counter objects, which cover a room that has snapshots but no
+// update log.
```
pyshx: same as above

server/websocket-go/internal/gcs/rooms.go
```diff
@@ -0,0 +1,97 @@
...
+	for r := range seen {
+		out = append(out, r)
+	}
+	return out, nil
+}
+
+var (
+	_ persistence.RoomLister                   = (*Adapter)(nil)
+	_ persistence.SnapshotVersionedPersistence = (*Adapter)(nil)
+)
```
pyshx: Please move all the compile-time interface assertion into the tests files and production file free of declarations that exist only for the type checker, and it still fails the build — the test binary won't compile.

server/websocket-go/internal/gcs/rooms_test.go
```diff
@@ -0,0 +1,72 @@
...
+// throughout this package (New(Options{..., Phase2: true})) rather than
+// introducing a new one, and covers: a room with an update log, a
+// snapshot-only room (no update log — the case most likely to be missed),
+// and room names containing a colon and a non-ASCII character, asserting the
+// ORIGINAL name comes back unencoded.
+//
+// A room name containing '/' is intentionally NOT covered here: Phase 2 uses
+// the room id raw as a "{room}/" path prefix, and ValidateDocIDForPrefix
+// rejects any id containing '/' for that reason. The Phase-1 conformance run
+// above already covers the slash case.
```
pyshx: Please remove this verbose.

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+)
+
+// SaveSnapshot stores state as a new labelled snapshot and returns its id.
+//
+// state is treated as an opaque, self-contained blob — exactly like the ygo
+// memory and file backends treat it — and is never decoded as a CRDT update. It
+// is only brotli-compressed for storage; GetSnapshotState decompresses and
+// returns it byte-for-byte. This matters because SnapshotStore's contract (and
+// its conformance suite) allows any non-empty byte string as state, not only a
+// valid Yjs update.
```
pyshx: Same as above.

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+	"strings"
+	"unicode/utf8"
+
+	"github.com/reearth/ygo/persistence"
+)
+
+// Custom-metadata keys on a snapver object. Label and the exact state length are
+// stored here so ListSnapshots is a single listAttrs call that never reads a
+// payload. attrs.Size is the brotli-COMPRESSED size and is deliberately unused
+// for SnapshotInfo.Size.
```
pyshx: Same as above.

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+				return nil
+			}
+			if !errors.Is(perr, errObjectExists) || attempt >= maxSnapshotIDAttempts {
+				return perr
+			}
+			// The id was already taken, so the counter was behind the objects that
+			// actually exist — a lock lease that expired mid-allocation, or a counter
+			// restored from an older state. Re-derive the floor from reality rather
+			// than trusting the counter again, and retry. The write-once precondition
+			// is what makes this safe: the losing writer never destroys the winner.
```
pyshx: The code is self-explanatory. Please make it either a one liner or remove it.

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+	})
+	if err != nil {
+		return 0, err
+	}
+	return id, nil
+}
+
+// maxSnapshotIDAttempts bounds SaveSnapshot's retry when an allocated id turns
+// out to be taken. Each retry re-derives the floor from stored objects, so more
+// than a couple of rounds means something is badly wrong rather than racing.
```
pyshx: Remove this.

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+const maxSnapshotIDAttempts = 3
+
+// maxLabelBytes bounds the label written into GCS custom object metadata. Real
+// GCS caps total custom metadata at 8 KiB per object; this leaves ample room for
+// the size key alongside it. The HTTP layer also bounds the label, but this is
+// the guard that matters: SaveSnapshot has more than one caller (ygo
+// auto-versioning is the other), and the storage layer is where the constraint
+// actually lives.
```
pyshx: One liner

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+const maxLabelBytes = 1024
+
+// ErrInvalidSnapshotLabel reports a label that cannot be stored in object
+// metadata.
```
pyshx, line 87: on liner

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+var ErrInvalidSnapshotLabel = errors.New("gcs: invalid snapshot label")
+
+// validateSnapshotLabel rejects labels GCS cannot round-trip.
+//
+// Both failures are silent without this check. An oversized label surfaces as an
+// opaque GCS 400 *after* the id counter has been consumed, so each attempt burns
+// an id and the user just sees "save failed". Invalid UTF-8 is worse: it is
+// accepted and mangled, so the label read back is not the label written.
```
pyshx: one liner

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+//
+// The counter object is a fast path, NOT the source of truth. When it is missing
+// or unparseable this falls back to deriving the floor from the snapshot objects
+// that exist. Trusting the counter alone is a data-loss bug: SaveSnapshot would
+// restart at id 1 and overwrite live snapshot 1, destroying its payload, label
+// and timestamp. That is routinely reachable, because Delete removes objects one
+// at a time and a cancelled request can take the counter while snapshots remain.
+//
+// Caller holds gcsDocLockKey(d), so this read-modify-write is race-free against
+// concurrent SaveSnapshot calls for the same room.
```
pyshx: same as above.

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return 0, err
+	}
+	return maxID + 1, nil
+}
+
+// maxSnapIDFromObjects returns the highest snapshot id actually present for the
+// room, or 0 when there are none. Both layouts are scanned when a fallback is
+// configured (Phase 2), so a Phase-2 allocation can never collide with a
+// legacy-root id that ListSnapshots still surfaces.
```
pyshx: Same as above.

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+				maxID = id
+			}
+		}
+	}
+	return maxID, nil
+}
+
+// snapshotLayouts returns the layouts whose snapshot objects are readable: the
+// primary, plus the legacy root when running Phase 2 so pre-cutover snapshots
+// stay visible. Order matters — the primary comes first so it wins de-duplication.
```
pyshx: Same as above

server/websocket-go/internal/gcs/snapshots.go
```diff
@@ -0,0 +1,267 @@
...
+	d := DocID(room)
+	for _, l := range a.snapshotLayouts() {
+		if err := a.store.delete(ctx, l.SnapVersionName(d, id)); err != nil && err != errNotFound {
+			return err
+		}
+	}
+	return nil
+}
+
+var _ persistence.SnapshotStore = (*Adapter)(nil)
```
pyshx: move to the test file

server/websocket-go/internal/server/persistence.go
```diff
@@ -15,10 +15,17 @@ import (
...
+	// Bound retained snapshots. LegacyAdapter.SaveVersion applies this after each
+	// save; it is a different axis from KeepVersions (the update log).
+	adapter.KeepSnapshots = cfg.KeepSnapshots
 	s := &Server{
 		cfg: cfg,
 		ws:  ygws.NewServerWithPersistence(adapter),
 		log: slog.Default(),
+		// Retained so the retention wiring is assertable: ygo keeps the adapter in
+		// an unexported field with no getter, so without this a test cannot tell
+		// KeepSnapshots was ever applied.
```
pyshx: Please remove

pyshx (changes requested):

bnimit:
## Why `NamedSnapshot` rather than extending `ProjectSnapshot`

Fair question, and the instinct behind it is right: this schema already has four near-identical snapshot-shaped types (`PreviewSnapshot`, `ProjectDocument`, `ProjectSnapshot`, `ProjectSnapshotMetadata`), so adding a fifth deserves justification. Two concrete reasons, and then what I think the real answer to the redundancy is.

### 1. They are keyed by different, non-interchangeable ids

```graphql
type ProjectSnapshot {          type NamedSnapshot {
  timestamp: DateTime!            id: Int!        # per-room snapshot counter
  updates: [Int!]!                label: String!
  version: Int!   # CRDT clock    timestamp: DateTime!
}                                 size: FileSize!
                                }
```

`ProjectSnapshot` has no identity of its own. It is addressed purely by `version`, a CRDT update-log clock. `NamedSnapshot.id` is a per-room counter that starts at 1 and has no relationship to any clock.

That distinction is load-bearing rather than pedantic. `rollbackProject(projectId:, version:)` reaches `PruneAfter`, which **deletes every update above the target clock**. Passing a snapshot id where a clock is expected prunes at an unrelated point and destroys history. A regression of exactly that kind was caught during review of this branch, which is why the UI rows are currently read-only (#2332).

Merging the two would put `id` and `version` on the same object, both `Int`, one of which silently destroys data. Keeping them as separate types is what makes that mistake hard to make by accident. That is also why `NamedSnapshot` deliberately does **not** carry a `version` field.

### 2. `ProjectSnapshot` carries the document payload; the list must not

`ProjectSnapshot.updates: [Int!]!` is non-null and is the full CRDT update vector, i.e. a complete copy of the document.

`projectSnapshots` is a **list**. Reusing that type would mean either returning N full document copies to render a version panel, or returning an empty `updates` array on every row and making a non-null field meaningless.

The storage design depends on the split too. Label and size live in GCS custom object metadata specifically so listing is one `listAttrs` call that never reads a payload:

> `ListSnapshots` returns snapshot metadata newest-first. It reads only object attributes (name + custom metadata), never a state blob.

And the UI only selects `id`, `label`, `timestamp`, `size` — never `updates`.

### On extending `ProjectSnapshotMetadata` instead

This is the stronger form of the question, since that type is already "metadata without updates". It still does not fit: it is the return type of `projectHistory`, which lists update-log clocks. Adding `id`/`label`/`size` there would give every `projectHistory` row three fields that are meaningless for an update-log entry, with no way for a client to tell which rows are which, and it would reintroduce problem 1 by seating `id` next to `version`.

### What I think the real answer is

The redundancy you are pointing at is genuine, but it predates this PR: `PreviewSnapshot`, `ProjectDocument` and `ProjectSnapshot` are three barely-distinguishable shapes already. The clean fix is to consolidate that family and give the two id spaces distinct scalar types so a mix-up is a compile error rather than a runtime data loss. That is a breaking schema change touching the UI, so I would rather not fold it into this PR.

Happy to file it as a follow-up alongside #2332, which is where snapshot ids first become client-addressable and the distinction starts to matter in practice. Say the word and I will open it.

bnimit:
## Comment trimming and interface assertions — done in e3402ff52

Applied both asks across the whole branch rather than only at the flagged lines, so the same feedback should not need repeating.

**Comments.** Cut to one or two lines throughout: `config`, `kv`, `layout`, `rooms`, `snapshots`, the http layer, the GraphQL schema, the API client and the tests. Where a comment only restated the code it is gone. Where it recorded something not visible in the code, it is now a single line — for example why the snapshot id counter is not the source of truth, why the snapshot state response is not `DocumentResponse`, and why an unsupported backend answers 501 rather than 404.

**Interface assertions.** Moved out of every `internal/gcs` production file into `snapshots_test.go`. That includes the pre-existing block in `persistence.go`, since leaving it would have half-done the request. Those production files now carry none.

**One thing worth flagging.** My first pass at trimming was scripted, and it truncated fourteen comment blocks mid-sentence, which is worse than the verbosity it replaced. I caught it by scanning for comment blocks whose last line has no terminal punctuation, then rewrote each by hand as a complete short comment. Mentioning it because a mechanical fix for a style problem turned into a legibility problem, and it is the kind of thing worth a second look during review.

I also dropped the long id-space warning from the `NamedSnapshot` schema comment. That hazard is now tracked in #2344, which proposes distinct scalar types so the mistake becomes a compile error rather than a comment nobody reads.

Verified: build, full test suites, and `golangci-lint` clean on both modules using CI's `--new-from-rev=origin/main` invocation, plus `make gql` reports no drift.

Resolving the individual threads now. Please reopen any where the remaining one-liner is still more than you want.

bnimit:
Renamed to `snapshotNumber` in e101bffa6.

```graphql
type NamedSnapshot {
  snapshotNumber: Int!
  label: String!
  timestamp: DateTime!
  size: FileSize!
}
```

Agreed `id` was wrong. It reads as a globally-unique identifier, and ygo's contract is explicit that the value is monotonically increasing *within* a room and **not** globally unique, so the same number names a different snapshot in another room. `snapshotNumber` says what it actually is.

For the record on the earlier question, since it came up twice: `version` counts **edits** (the CRDT update-log clock, `clock = last + 1` per persisted update) while this counts **saved snapshots**, from a separate counter over a separate object family. Snapshot number 3 may hold the document as of version 487. There is also no mapping between them, because a snapshot does not record the clock it was taken at — the stored metadata is label and size only. That is why a merged type could not populate `version` for snapshot rows even if we wanted it to.

Schema, generated code and both resolvers updated. `make gql` reports no drift; build, tests and `golangci-lint` clean using CI's `--new-from-rev` invocation.

The UI branch selects this field, so it takes the same rename. It is unpushed and already due a rebase once this merges, so that folds into the same pass rather than a separate round trip.

Still open from the discussion: whether to collapse `label`/`size` into `ProjectSnapshotMetadata`. You are right that those two collapse cleanly; the blocker is `version`, which would have to become nullable (it is `Int!` today and `projectHistory` consumers assume a value). I would rather do that in #2344 alongside the id-type split, but happy to do it here if you prefer — roughly an hour with the UI null-checks. The one thing I would avoid is a non-null `version` defaulting to 0 on snapshot rows, since 0 is the value that prunes the whole log.

server/api/gql/document.graphql
```diff
@@ -28,6 +28,15 @@ type ProjectSnapshotMetadata {
   version: Int!
 }
 
+# A labelled snapshot in a project's version history. snapshotNumber counts saved
+# snapshots per room; it is not an update-log version. See reearth-flow#2344.
```
pyshx: No need to mention the PR(#2344) and please make this a one liner

server/api/gql/document.graphql
```diff
@@ -44,4 +54,5 @@ extend type Query {
   latestProjectSnapshot(projectId: ID!): ProjectDocument
   projectHistory(projectId: ID!): [ProjectSnapshotMetadata!]!
   projectSnapshot(projectId: ID!, version: Int!): ProjectSnapshot!
+  projectSnapshots(projectId: ID!): [NamedSnapshot!]!
```
pyshx: This is a little ambigous query, I think its better to rename it to `projectNamedSnapshots`

server/api/internal/usecase/interactor/previewschema_test.go
```diff
@@ -52,6 +52,12 @@ func (w *fakeWebsocket) GetHistoryByVersion(context.Context, string, int) (*webs
 func (w *fakeWebsocket) GetHistoryMetadata(context.Context, string) ([]*websocket.HistoryMetadata, error) {
 	return nil, nil
 }
+func (w *fakeWebsocket) GetSnapshots(context.Context, string) ([]*websocket.SnapshotMetadata, error) {
```
pyshx: Same as above, Its better to find a better name for this.

server/websocket-go/internal/gcs/rooms.go
```diff
@@ -0,0 +1,81 @@
...
+	"unicode/utf8"
+)
+
+// ListRooms enumerates every room the bucket holds data for.
+//
+// NOTE: not wired yet; admin cleanup still enumerates live rooms only. See
+// reearth/reearth-flow#2333, which also lists two bugs to fix before wiring.
+//
+// Phase 2 is a prefix listing. Phase 1 recovers rooms from the OID index objects
+// plus the snapshot counters, which cover snapshot-only rooms.
```
pyshx: Please add a TODO instead of "Phase" that might not be relevant to the code itself

pyshx (approved):

bnimit:
All four addressed in b5d8d3410.

- **`projectSnapshots` → `projectNamedSnapshots`.** Agreed the old name was ambiguous: this surface has two kinds of snapshot, and it read as though it returned the same thing as the existing `projectSnapshot` query. The client method follows for symmetry with `SaveNamedSnapshot`, so `GetSnapshots` → `GetNamedSnapshots` — that was the "same as above" on `previewschema_test.go`, let me know if you meant something else there.
- **Dropped the #2344 reference** from the `NamedSnapshot` comment and cut it to one line.
- **`ListRooms` now carries a `TODO(reearth/reearth-flow#2333)`** instead of the Phase 1/Phase 2 wording. Good call — the phase naming describes a migration state rather than the code, so it would have aged badly.

Schema, generated code, resolver, interface, client and every test updated; no old names remain. `make gql` reports no drift, and build, tests and `golangci-lint` are clean on both modules.

Your approval is intact after the push — the repo ruleset has `dismiss_stale_reviews_on_push: false` — so this should be mergeable without another look unless you want one.

Two notes for after the merge:

- The UI branch selects this query and now inherits two renames (`snapshotNumber` and `projectNamedSnapshots`). It is unpushed and already due a rebase, so both land in one pass.
- #2341 (the authorization fix) touches `interactor/websocket.go` too. Whichever of the two merges second needs `GetNamedSnapshots` and `SaveNamedSnapshot` added to the authorizing wrapper, since this PR extends `interfaces.WebsocketClient`. Happy to do that rebase as soon as one lands.

PR #2339, opened 2026-08-04
feat(engine): port GeoJSON writers to new geometry

n4to4:
# Overview

Ports `GeoJSON Writer` (sink) and `Feature GeoJSON Writer` (processor) to the new geometry world, and makes the legacy GeoJSON 2008 `crs` member opt-in via a new `writeCrs` parameter.

## What I've done

- Added the `Feature -> GeoJSON` direction to `conversion/geojson_next.rs` and removed the `not(new-geometry)` gates from both writers, so they work in either geometry world. Both directions now share one axis-order invariant: GeoJSON positions are always `(easting/longitude, northing/latitude[, height])`, so a north-first CRS frame has its horizontal pair swapped on the way out.
- Added `writeCrs` (default `false`) to both actions. When enabled, the `crs` member is a named EPSG CRS if one code covers every written coordinate, and an explicit `null` otherwise — an absent member would claim WGS84 for coordinates that are not in it. `writeCrs: true` is set on the two error-geometry outputs in `quality-check/plateau6/02-bldg/building_checks.yml`, the only in-repo nodes that need it.
- Moved `feature/geojson_writer.rs` to `feature/writer/geojson.rs`.

## How I tested

Added unit tests for the write direction and the `crs` member, running in both geometry worlds.

- Writing `"crs": null` rather than omitting the member when the CRS is unknown or mixed.
- The collection folding rules in `combine` / `fold_into_multi`, especially requiring a uniform coordinate frame.

## Memo

- **CRS**: a named CRS only when one EPSG code covers *every written* coordinate — coverage accumulates during the write, so a dropped geometry contributes nothing. Anything else (mixed codes, a frame naming no CRS, nothing written) is an explicit `"crs": null`, since an omitted member would claim the GeoJSON 2008 default, WGS84 lon/lat. That member is a GeoJSON 2008 extension RFC 7946 dropped, hence opt-in.
- **Axis order**: RFC 7946 3.1.1 fixes positions to (easting/longitude, northing/latitude), while a `Crs` frame stores axes in the authority's declared order — so writing swaps the horizontal pair back. `Euclidean` and `Tangent` frames are written as stored.
- **Solid / Csg / PointCloud** have no GeoJSON counterpart: dropped with a warning where they appear so they don't discard their siblings, though a container left with nothing writable is an error. Meshes write as their faces.
- **Collections**: a `GeometryCollection` stays one GeoJSON `GeometryCollection` on one feature, at any depth — not the old world's expansion into one feature per member. Reading builds one `Feature` from one GeoJSON feature, so writing one back makes the two directions inverses, ids and all. Parts fold into a `Multi*` only when they share a kind and a frame — folding across frames would put two reference systems under one `crs`.

### DIFF
--- a/engine/runtime/types/src/conversion/geojson_next.rs
+++ b/engine/runtime/types/src/conversion/geojson_next.rs
@@ -7,14 +7,16 @@
 //! while a CRS frame stores its axes in the order the CRS authority declares, so
 //! reading swaps the horizontal pair and writing swaps it back.
 
-use std::sync::Arc;
+use std::collections::HashSet;
+use std::sync::{Arc, Mutex, OnceLock};
 
+use itertools::Itertools;
 use reearth_flow_geometry::types::conversion::{is_2d_geojson_value, is_3d_geojson_value};
 use reearth_flow_geometry::{
     collection::{Collection2D, Collection3D},
     coordinate::{CoordinateFrame, EpsgCode},
     line_string::{LineString2D, LineString3D},
-    ops::Split,
+    ops::{Split, UnsupportedOperation},
     point::{Point2D, Point3D},
     polygon::{Polygon2D, Polygon3D},
     Euclidean2DGeometry, Euclidean3DGeometry, Geometry, GeometryCollection,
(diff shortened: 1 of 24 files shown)


engine/runtime/types/src/conversion/geojson_next.rs
```diff
@@ -196,11 +206,530 @@ fn polygon_3d(rings: &[Vec<Vec<f64>>]) -> Polygon3D {
...
+}
+
+fn write_2d(geometry: &Euclidean2DGeometry) -> Result<WrittenGeometry, Unwritable> {
+    use Euclidean2DGeometry::*;
+    match geometry {
+        Point(p) => Ok(point(p.frame(), p.position())),
+        LineString(l) => Ok(curve(l.frame(), l.coords())),
+        Polygon(p) => Ok(area(p.frame(), p.exterior(), p.interiors())),
+        PolygonMesh(m) => write_faces((**m).clone()),
+        TriangularMesh(m) => write_faces((**m).clone()),
```
shunski, line 363: Clone should not always be necessary as `split` only actively deconstruct it. But this would rarelly bites as GeoJson output is usually small, so can be left as is.

shunski (approved):
Approved with a minor comment

PR #2341, opened 2026-08-04
fix(api): authorize the document/websocket GraphQL operations

### LINKED ISSUE #2337: Document/websocket GraphQL operations have no workspace authorization check

bnimit:
Closes #2337.

## The problem

Every operation on the document/websocket GraphQL surface takes a `projectId` and passed it straight to the websocket HTTP client. That client knows nothing about projects or workspaces, it just forwards. So a caller only had to be **authenticated**, not authorized, and any signed-in user could act on any project in any workspace:

- read another workspace's document state and version history
- call `rollbackProject`, which reaches `PruneAfter` and deletes every update above the target clock, making it a **destructive cross-tenant operation**
- call `importProject` / `copyProject`, writing into projects they do not own

Snapshot labels are user-authored and often descriptive, so even the read path leaks business information without fetching any state.

The cause was a single line of wiring. `interfaces.Container.Websocket` held the bare infrastructure client, while every other `Container` entry is a permission-checking interactor, so no resolver on this surface had a permission layer to inherit from.

## The fix

`interactor.Websocket` now decorates the client: resolve the project to its owning workspace, check `rbac.ResourceProject`, then delegate.

| Operation | Action |
|---|---|
| `latestProjectSnapshot`, `projectHistory`, `projectSnapshot` | `ActionRead` |
| `previewSnapshot` | `ActionRead` |
| `rollbackProject`, `saveSnapshot`, `importProject` | `ActionEdit` |
| `copyProject` | `ActionEdit` on destination + `ActionRead` on source |
| `DeleteDocument` | `ActionDelete` |

Three decisions worth reviewing:

- **`previewSnapshot` is `ActionRead`, not `ActionEdit`.** Despite the underlying method being named `CreateSnapshot`, the websocket server only materializes state at a version and returns it. Requiring edit would wrongly deny read-only users a preview.
- **`copyProject` checks both ends.** Checking only the destination would let a caller pull another workspace's document into a project they legitimately own.
- **Authorization fails closed at every step.** An unparseable id, a missing project, or a lookup error all deny rather than fall through, so a malformed id cannot be used to skip the check.

**Internal callers stay on the raw client, deliberately.** `ProjectDeleter` already checks `ActionDelete` on the workspace, and it cleans up the document *before* removing the project row. Routing it through here would be redundant, and would break if that ordering ever changed.

This also gives `interactor.Websocket` a purpose. It was dead code slated for deletion in #2334: unreferenced, and carrying method names that did not match the interface it was supposedly for. Rather than delete it and add a new type, it becomes the authorization layer it should always have been. **#2334 can be closed once this merges.**

## Testing

Unit tests cover all nine operations in a table, asserting each demands the right action, scopes the check to the addressed project's workspace, and never reaches the client when denied. Verified against three mutants:

| Mutant | Result |
|---|---|
| Authorization is a no-op (the actual pre-fix state) | 18 subtests fail |
| Check runs but ignores the workspace (the cross-tenant hole) | 9 subtests fail |
| `copyProject` checks only the destination | fails |

The e2e test earns its place by covering what the unit tests cannot: that the container actually points at this interactor. Re-pointing it at the raw client leaves every unit test green. It is memory-backed, so it runs without a database.

Two corrections were needed to give that test teeth, and both are worth knowing about:

1. Asserting only "a GraphQL error occurred" **passes under the very bypass it targets**, because an unreachable websocket server errors too. I confirmed that by mutation. It now asserts the message says `operation denied` and does **not** say `server returned`.
2. That tightening then exposed a second problem: the `importProject` row was failing `Bytes` scalar coercion before reaching the resolver, so it had been asserting nothing at all.

Also worth flagging for a follow-up: the pre-existing `TestDocumentOperations` is not coverage of this surface. Its interceptor answers the GraphQL request itself with canned JSON, so it never reaches the resolvers, the container, or the client.

## Verification

`go build ./...` clean, `golangci-lint` clean on touched packages, unit and new e2e tests pass. The Mongo-backed e2e tests were not run locally (no database available); CI covers them.

### DIFF
--- a/server/api/e2e/gql_document_permission_test.go
+++ b/server/api/e2e/gql_document_permission_test.go
@@ -1,0 +1,185 @@
+package e2e
+
+import (
+	"context"
+	"net/http"
+	"testing"
+
+	"github.com/gavv/httpexpect/v2"
+	usermockrepo "github.com/reearth/reearth-accounts/server/pkg/gqlclient/user/mockrepo"
+	accountsuser "github.com/reearth/reearth-accounts/server/pkg/user"
+	"github.com/reearth/reearth-flow/api/internal/app/config"
+	"github.com/reearth/reearth-flow/api/internal/testutil/factory"
+	"github.com/reearth/reearth-flow/api/internal/usecase/repo"
+	"github.com/reearth/reearth-flow/api/pkg/project"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+// The document/websocket operations used to be the one GraphQL surface with no
+// authorization: interfaces.Container.Websocket held the bare HTTP client, so any
+// authenticated user could act on any project in any workspace. rollbackProject
+// was the worst of it, because it prunes a project's update log.
+//
+// These tests pin the WIRING, which the interactor unit tests cannot reach. Those
+// prove the interactor denies; only booting the real server proves
+// Container.Websocket actually points at the interactor and not the raw client.
+// Re-pointing it would leave every unit test green.
+//
+// Memory-backed on purpose (useMongo = false) so this runs anywhere rather than
+// only where a database is available.
+//
+// Note the pre-existing TestDocumentOperations is not comparable coverage: its
+// interceptor answers the GraphQL request itself with canned JSON, so it never
+// reaches the resolvers, the container, or the client.
+
+// startDocumentPermissionServer boots the real server with permissions either
+// enforced or granted, and returns the expect client plus the project repo.
+func startDocumentPermissionServer(t *testing.T, allowPermission bool) (*httpexpect.Expect, repo.Project) {
+	t.Helper()
+	ctrl := gomock.NewController(t)
+
+	workspace := factory.NewWorkspace()
+	userEntity := factory.NewUser(func(b *accountsuser.Builder) {
+		b.Workspace(workspace.ID())
+		b.Auths([]accountsuser.Auth{{Provider: "auth0", Sub: "auth0|doc-perm-test"}})
+	})
+
+	mockUserRepo := usermockrepo.NewMockRepo(ctrl)
+	// AnyTimes: the table below issues one request per operation.
+	mockUserRepo.EXPECT().FindMe(gomock.Any()).Return(userEntity, nil).AnyTimes()
+
+	cfg := &config.Config{
+		Origins:         []string{"https://example.com"},
+		AuthSrv:         config.AuthSrvConfig{Disabled: true},
+		AccountsApiHost: "http://localhost:8080",
+	}
+	exp, repos, _ := StartServerAndRepos(t, cfg, false, allowPermission, &TestMocks{UserRepo: mockUserRepo})
+	return exp, repos.Project
+}
+
+// documentOps is every operation the schema exposes. Table-driven so a new
+// operation added without authorization shows up as a missing row rather than
+// passing unnoticed.
+var documentOps = []struct {
+	name  string
+	query string
+	vars  func(projectID string) map[string]any
+}{
+	{
+		name:  "latestProjectSnapshot",
+		query: `query($projectId: ID!) { latestProjectSnapshot(projectId: $projectId) { version } }`,
+		vars:  func(p string) map[string]any { return map[string]any{"projectId": p} },
+	},
+	{
+		name:  "projectHistory",
+		query: `query($projectId: ID!) { projectHistory(projectId: $projectId) { version } }`,
+		vars:  func(p string) map[string]any { return map[string]any{"projectId": p} },
+	},
+	{
+		name:  "projectSnapshot",
+		query: `query($projectId: ID!, $version: Int!) { projectSnapshot(projectId: $projectId, version: $version) { version } }`,
+		vars: func(p string) map[string]any {
+			return map[string]any{"projectId": p, "version": 1}
+		},
+	},
+	{
+		name:  "rollbackProject",
+		query: `mutation($projectId: ID!, $version: Int!) { rollbackProject(projectId: $projectId, version: $version) { version } }`,
+		vars: func(p string) map[string]any {
+			return map[string]any{"projectId": p, "version": 1}
+		},
+	},
+	{
+		name:  "saveSnapshot",
+		query: `mutation($projectId: ID!) { saveSnapshot(projectId: $projectId) }`,
+		vars:  func(p string) map[string]any { return map[string]any{"projectId": p} },
+	},
+	{
+		name:  "previewSnapshot",
+		query: `mutation($projectId: ID!, $version: Int!) { previewSnapshot(projectId: $projectId, version: $version) { version } }`,
+		vars: func(p string) map[string]any {
+			return map[string]any{"projectId": p, "version": 1}
+		},
+	},
+	{
+		name:  "importProject",
+		query: `mutation($projectId: ID!, $data: Bytes!) { importProject(projectId: $projectId, data: $data) }`,
+		vars: func(p string) map[string]any {
+			// Bytes is a number array (Uint8Array), NOT base64: a string here fails
+			// scalar coercion before the resolver runs, so the row would assert nothing.
+			return map[string]any{"projectId": p, "data": []int{1, 2, 3}}
+		},
+	},
+}
+
+// TestDocumentOperations_DeniedWithoutPermission: with the permission checker
+// refusing, every operation must return a GraphQL error.
+//
+// This is what pins the authorizing wrapper in place. If Container.Websocket were
+// re-pointed at the raw client, the request would instead reach a websocket server
+// that is not running, which is a different failure and a real bypass.
+func TestDocumentOperations_DeniedWithoutPermission(t *testing.T) {
+	e, projectRepo := startDocumentPermissionServer(t, false)
+
+	prj := project.New().NewID().Workspace(project.NewWorkspaceID()).MustBuild()
+	require.NoError(t, projectRepo.Save(context.Background(), prj))
+
+	for _, op := range documentOps {
+		t.Run(op.name, func(t *testing.T) {
+			res := e.POST("/api/graphql").
+				WithHeader("Origin", "https://example.com").
+				WithHeader("authorization", "Bearer test").
+				WithHeader("Content-Type", "application/json").
+				WithJSON(map[string]any{
+					"query":     op.query,
+					"variables": op.vars(prj.ID().String()),
+				}).
+				Expect().
+				Status(http.StatusOK).
+				JSON().Object()
+
+			// Assert the SPECIFIC error. "an error occurred" is not enough: with the
+			// bare client wired in, the request reaches a websocket server that is not
+			// running and fails with `server returned non-200 status: 404`, so a test
+			// that only checks for the presence of errors passes under the very bypass
+			// it is meant to catch. (Verified: it does.) Only the message distinguishes
+			// authorization from a transport failure.
+			res.ContainsKey("errors")
+			msg := res.Value("errors").Array().Value(0).Object().Value("message").String()
+			msg.Contains("operation denied")
+			msg.NotContains("server returned")
+		})
+	}
+}
+
+// TestDocumentOperations_UnknownProjectIsDenied: authorization resolves the
+// project to a workspace, so an id that does not resolve must fail closed rather
+// than skip the check. Permissions are GRANTED here, so the only thing that can
+// produce an error is the resolution step refusing.
+func TestDocumentOperations_UnknownProjectIsDenied(t *testing.T) {
+	e, _ := startDocumentPermissionServer(t, true)
+
+	for _, projectID := range []string{"not-a-valid-id", project.NewID().String()} {
+		t.Run(projectID, func(t *testing.T) {
+			res := e.POST("/api/graphql").
+				WithHeader("Origin", "https://example.com").
+				WithHeader("authorization", "Bearer test").
+				WithHeader("Content-Type", "application/json").
+				WithJSON(map[string]any{
+					"query":     `query($projectId: ID!) { projectHistory(projectId: $projectId) { version } }`,
+					"variables": map[string]any{"projectId": projectID},
+				}).
+				Expect().
+				Status(http.StatusOK).
+				JSON().Object()
+
+			// Same reasoning as above, via the negative: permissions are granted here,
+			// so the request must be stopped by project resolution BEFORE the client is
+			// reached. "server returned" in the message would mean it went through.
+			res.ContainsKey("errors")
+			msg := res.Value("errors").Array().Value(0).Object().Value("message").String()
+			msg.NotContains("server returned")
+		})
+	}
+}
--- a/server/api/e2e/gql_document_permission_test.go
+++ b/server/api/e2e/gql_document_permission_test.go
@@ -58,74 +58,23 @@ func startDocumentPermissionServer(t *testing.T, allowPermission bool) (*httpexp
 	return exp, repos.Project
 }
 
-// documentOps is every operation document.graphql exposes: three queries
-// (latestProjectSnapshot, projectHistory, projectSnapshot) and five mutations
-// (rollbackProject, saveSnapshot, previewSnapshot, importProject, copyProject).
-// Table-driven so a new operation added without authorization shows up as a
-// missing row rather than passing unnoticed.
-var documentOps = []struct {
-	// vars first: keeping the pointer-bearing field at the front shortens the
-	// range the GC has to scan (govet fieldalignment).
-	//
-	// It takes a second project id so copyProject, the one operation addressing
-	// two projects, can be covered by the same table rather than sitting outside
-	// it where a coverage gap is easy to miss.
-	vars  func(projectID, otherProjectID string) map[string]any
-	name  string
-	query string
-}{
-	{
-		name:  "latestProjectSnapshot",
-		query: `query($projectId: ID!) { latestProjectSnapshot(projectId: $projectId) { version } }`,
-		vars:  func(p, _ string) map[string]any { return map[string]any{"projectId": p} },
-	},
-	{
-		name:  "projectHistory",
-		query: `query($projectId: ID!) { projectHistory(projectId: $projectId) { version } }`,
-		vars:  func(p, _ string) map[string]any { return map[string]any{"projectId": p} },
-	},
-	{
-		name:  "projectSnapshot",
-		query: `query($projectId: ID!, $version: Int!) { projectSnapshot(projectId: $projectId, version: $version) { version } }`,
-		vars: func(p, _ string) map[string]any {
-			return map[string]any{"projectId": p, "version": 1}
-		},
-	},
-	{
-		name:  "rollbackProject",
-		query: `mutation($projectId: ID!, $version: Int!) { rollbackProject(projectId: $projectId, version: $version) { version } }`,
-		vars: func(p, _ string) map[string]any {
-			return map[string]any{"projectId": p, "version": 1}
-		},
-	},
-	{
-		name:  "saveSnapshot",
-		query: `mutation($projectId: ID!) { saveSnapshot(projectId: $projectId) }`,
-		vars:  func(p, _ string) map[string]any { return map[string]any{"projectId": p} },
-	},
-	{
-		name:  "previewSnapshot",
-		query: `mutation($projectId: ID!, $version: Int!) { previewSnapshot(projectId: $projectId, version: $version) { version } }`,
-		vars: func(p, _ string) map[string]any {
-			return map[string]any{"projectId": p, "version": 1}
-		},
-	},
-	{
-		name:  "importProject",
-		query: `mutation($projectId: ID!, $data: Bytes!) { importProject(projectId: $projectId, data: $data) }`,
-		vars: func(p, _ string) map[string]any {
-			// Bytes is a number array (Uint8Array), NOT base64: a string here fails
-			// scalar coercion before the resolver runs, so the row would assert nothing.
-			return map[string]any{"projectId": p, "data": []int{1, 2, 3}}
-		},
-	},
-	{
-		name:  "copyProject",
-		query: `mutation($projectId: ID!, $source: ID!) { copyProject(projectId: $projectId, source: $source) }`,
-		vars: func(p, other string) map[string]any {
-			return map[string]any{"projectId": p, "source": other}
-		},
-	},
+// assertDenied posts one operation and requires the SPECIFIC authorization
+// error. "an error occurred" is not enough: with the bare client wired in the
+// request reaches a websocket server that is not running and fails with
+// `server returned non-200 status: 404`, which passes that weaker check.
+func assertDenied(t *testing.T, e *httpexpect.Expect, name, query string, vars map[string]any) {
+	t.Helper()
+	res := e.POST("/api/graphql").
+		WithHeader("Origin", "https://example.com").
+		WithHeader("authorization", "Bearer test").
+		WithHeader("Content-Type", "application/json").
+		WithJSON(map[string]any{"query": query, "variables": vars}).
+		Expect().Status(http.StatusOK).JSON().Object()
+
+	res.ContainsKey("errors")
+	msg := res.Value("errors").Array().Value(0).Object().Value("message").String()
+	msg.Contains("operation denied")
+	msg.NotContains("server returned")
 }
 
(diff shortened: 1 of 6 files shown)


bnimit:
## Merge-order note: interacts with #2338

Both PRs are backend-only and touch no UI, but they both modify `server/api/internal/usecase/interactor/websocket.go`, and the second one to merge **will not compile** without a small follow-up.

#2338 extends `interfaces.WebsocketClient` with two methods:

```go
GetSnapshots(ctx context.Context, docID string) ([]*websocket.SnapshotMetadata, error)
SaveNamedSnapshot(ctx context.Context, docID, label string) (*websocket.SnapshotMetadata, error)
```

The authorizing decorator here asserts `var _ interfaces.WebsocketClient = (*Websocket)(nil)` and implements neither, because neither exists on `main` yet. Nothing in this PR can pre-empt that: `pkg/websocket.SnapshotMetadata` is introduced by #2338.

### Suggested order: this PR first

It is small, self-contained, and a security fix, so it should not wait. When #2338 rebases, the two new methods get added to the decorator with the same treatment as their siblings, `ActionRead` for `GetSnapshots` and `ActionEdit` for `SaveNamedSnapshot`. That is a handful of lines in the file this PR already rewrites.

### If #2338 merges first, watch the conflict resolution

Resolving it by keeping the snapshot operations on the raw client would leave those two operations **unauthorized**, which is precisely the hole this PR closes. The two new methods must go through `authorize` like the rest. The table-driven unit test in `websocket_permission_test.go` is the place to add the rows, so a missing check shows up as a failure rather than an omission.

server/api/e2e/gql_document_permission_test.go
```diff
@@ -0,0 +1,185 @@
...
+	},
+	{
+		name:  "importProject",
+		query: `mutation($projectId: ID!, $data: Bytes!) { importProject(projectId: $projectId, data: $data) }`,
+		vars: func(p string) map[string]any {
+			// Bytes is a number array (Uint8Array), NOT base64: a string here fails
+			// scalar coercion before the resolver runs, so the row would assert nothing.
+			return map[string]any{"projectId": p, "data": []int{1, 2, 3}}
+		},
+	},
```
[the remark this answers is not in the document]
bnimit: Good catch, and this one was a real gap rather than a wording problem. `copyProject` writes into a project, so leaving it out of the wiring guard meant the guard did not cover the surface it claimed to.

Fixed by bringing it into the table rather than testing it alongside: the `vars` func now takes a second project id, so the one operation addressing two projects fits the same table instead of sitting outside it where a gap is easy to miss.

The two projects are seeded in **different workspaces** deliberately. Using one project for both ends would let a destination-only check look correct.

The table is now verifiably complete — every operation in `document.graphql` has a row:

```
$ comm -23 <schema ops> <table ops>
(empty)
```

Also re-ran the bare-client mutation afterwards to confirm both e2e tests still fail under the bypass now that `copyProject` is included, so the added row did not dilute anything.

Fixed in 8859d9fcc.

server/api/internal/usecase/interactor/websocket_permission_test.go
```diff
@@ -0,0 +1,230 @@
...
+	require.NoError(t, projectRepo.Save(ctx, prj))
+
+	client := &countingWSClient{}
+	rc := &recordingChecker{allow: allow}
+	return NewWebsocket(client, projectRepo, rc), client, rc, prj.ID().String(), accountsid.WorkspaceID(wsID)
+}
+
+// wsOps is every operation on the surface, with the action each must demand.
+// Table-driven so a newly added method without a permission check shows up as a
+// missing row rather than passing silently.
```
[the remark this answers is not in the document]
bnimit: Agreed, and the risk you name is the right one: a comment claiming the list is exhaustive is how a future method gets added with no permission check.

The comment now names both exclusions and points at the tests that cover them, plus **why** `CopyDocument` cannot simply join the table — which is the part a future contributor would otherwise have to rediscover:

> - `CopyDocument` addresses two projects, so it does not fit this signature, and it authorizes twice (edit on the destination, read on the source). Since `recordingChecker` keeps only the last call, a row here would assert `ActionRead` and quietly stop checking the destination.
>   See `TestWebsocket_CopyDocumentChecksBothProjects`.
> - `Close` is connection lifecycle, not a document operation, and carries no permission check at all. See `TestWebsocket_CloseNeedsNoPermission`.

That last-call detail matters: naively adding the row would have looked like more coverage while silently weakening it.

Worth noting the same review point applied to the e2e table for a different reason — there it was a genuine missing operation, not just an inaccurate comment. Both are fixed in 8859d9fcc.

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,158 @@ package interactor
...
+// Authorization belongs here rather than in the resolvers because this is the
+// use-case layer, and every other Container entry is already a
+// permission-checking interactor. Wiring the bare client into
+// Container.Websocket is what made this surface the exception.
+//
+// NOTE: internal callers are deliberately NOT routed through this type.
+// ProjectDeleter takes the raw client, because project deletion has already
+// checked ActionDelete on the workspace by the time it cleans up the document,
+// and re-resolving a project that is mid-deletion would be both redundant and
+// fragile.
```
pyshx: Please make it a liner of necessary or remove it.

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,158 @@ package interactor
...
-	client interfaces.WebsocketClient
+	client            interfaces.WebsocketClient
+	projectRepo       repo.Project
+	permissionChecker gateway.PermissionChecker
 }
 
-func NewWebsocket(client interfaces.WebsocketClient) *Websocket {
-	return &Websocket{client: client}
+// NewWebsocket wraps client so every operation is authorized against the owning
+// workspace of the project it addresses.
```
pyshx: Not needed.

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,158 @@ package interactor
...
 }
 
-func (i *Websocket) GetLatest(ctx context.Context, id string) (*ws.Document, error) {
-	return i.client.GetLatest(ctx, id)
+// authorize resolves docID to its project and checks action against that
+// project's workspace.
+//
+// It fails closed at every step. An unparseable id, a project that does not
+// exist, or a lookup error all deny rather than fall through — otherwise a
+// malformed id would be a way to skip the check entirely.
```
pyshx: same as above.

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,158 @@ package interactor
...
+	if err := i.authorize(ctx, docID, rbac.ActionRead); err != nil {
+		return nil, err
+	}
+	return i.client.GetHistoryMetadata(ctx, docID)
 }
 
-func (i *Websocket) FlushToGCS(ctx context.Context, id string) error {
-	return i.client.FlushToGCS(ctx, id)
+// Rollback is the destructive one: it prunes every update above the target
+// clock, so an unauthorized call permanently truncates a project's history.
```
pyshx: Make this warning a one liner.

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,158 @@ package interactor
...
+	if err := i.authorize(ctx, docID, rbac.ActionEdit); err != nil {
+		return err
+	}
+	return i.client.FlushToGCS(ctx, docID)
+}
+
+// CreateSnapshot backs previewSnapshot. Despite the name it does not write: the
+// websocket server materializes the state at the requested version and returns
+// it. ActionRead is therefore correct, and using ActionEdit would wrongly deny
+// read-only users a preview.
```
pyshx: One liner.

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,158 @@ package interactor
...
+		return nil, err
+	}
 	return i.client.CreateSnapshot(ctx, docID, version, name)
 }
 
-func (i *Websocket) CopyProject(ctx context.Context, id string, source string) error {
-	return i.client.CopyDocument(ctx, id, source)
+// CopyDocument reads source and writes docID, so it needs BOTH: edit on the
+// destination and read on the source. Checking only the destination would let a
+// caller pull another workspace's document into a project they own.
```
pyshx: One liner.

server/api/internal/usecase/interactor/websocket_permission_test.go
pyshx, line 1: Table driven test should be avoided unless the test involves initialisation of multiple initialisations.

:white_check_mark: Good: simple, enough to maintain
```
assert.Equal(t, 3, Sum(1, 2))
assert.Equal(t, 4, Sum(1, 3))
assert.Equal(t, 5, Sum(1, 4))
```

:x: Bad: Too Much
```
type args struct {
  a int
  b int
}
tests := []struct {
  name string
  args args
  want int
}{
  { name: "1+2", aegs:args{ a: 1, b: 2 }, want: 3 },
  { name: "1+3", aegs:args{ a: 1, b: 3 }, want: 4 },
  { name: "1+4", aegs:args{ a: 1, b: 4 }, want: 5 },
}
for _, tt := range tests {
  tt := tt
  t.Run(tt.name, func(t *testing.T) {
    assert.Equal(t, 3, Sum(tt.args.a, tt.args.b))
  })
}
```

pyshx (commented):
Thanks! Overall the PR looks good. Can you also make sure that the policy has also been kept in check for this change?

bnimit:
## Addressed in 8061c0ea6

**Comments** cut to one or two lines in `interactor/websocket.go`. Where a comment restated the code it is gone; where it recorded something the code does not show it is now one line — that `CreateSnapshot` does not write despite the name, that `CopyDocument` needs both ends, and that internal callers stay on the raw client because `ProjectDeleter` already checks `ActionDelete`.

**Table-driven tests** replaced with explicit sequential calls to two helpers, in both the unit and e2e tests. No struct slice, no loop, no `tt := tt`:

```go
assertChecks(t, "GetLatest", rbac.ActionRead, getLatest)
assertChecks(t, "Rollback", rbac.ActionEdit, rollback)
assertChecks(t, "DeleteDocument", rbac.ActionDelete, deleteDocument)
```

Agreed this reads better — the action each operation demands is now on the same line as the operation, rather than a field in a struct two screens away. It also keeps the one property the table was there for: every operation is still listed in one place, so adding a method without a permission check leaves a visible gap.

`CopyDocument` and `Close` stay out of both lists with their own tests. `CopyDocument` authorizes twice and `recordingChecker` keeps only the last call, so listing it would assert `ActionRead` and stop checking the destination.

**One thing I checked rather than assumed:** restructuring a guard can quietly weaken it, so I re-ran all three mutants afterwards. Bare client in the container still fails 2 e2e tests, a no-op `authorize` still fails 3 unit tests, and dropping the workspace argument still fails the action/workspace test. The flattening cost no coverage.

Build, tests and `golangci-lint` clean using CI's `--new-from-rev` invocation. Resolving the threads.

bnimit:
## Policy check — you were right, and it would have broken the feature

Thanks for asking. I had not kept the policy in check, and it was not a cosmetic miss.

### Two problems

`flow:project` declares `any`, `create`, `edit`, `delete`, `list`. It has **no `read` rule**, and this PR was the only code checking `ActionRead` against `ResourceProject`. Cerbos denies an action with no matching rule, so all five read operations — `projectHistory`, `latestProjectSnapshot`, `projectSnapshot`, `previewSnapshot`, and copy's source check — would have been denied for **every user including owners**. The version history panel would not have worked at all.

Separately, `flow:project`'s `edit` is maintainer and owner only, deliberately: editing a *project* means renaming and reconfiguring it. Using that action for document mutations would have denied **writers** the ability to save a version, roll back, import or copy, and writers are exactly the people doing the collaborative editing.

### The fix

Document operations get their own resource rather than widening `ResourceProject`:

```
flow:projectDocument
  read   -> reader, writer, maintainer, owner   (see the version history)
  edit   -> writer, maintainer, owner           (content work, as for assets)
  delete -> maintainer, owner                   (destroys all document data)
```

That follows the `writerMaintainerOwner` pattern already used for assets, parameters and nodes, on the reasoning that editing the canvas is project **content** while `ResourceProject.edit` is project **settings**. Adding writers to `ResourceProject.edit` would have let them rename and reconfigure projects too, which is a wider change than this PR needs.

### Why nothing caught it

Worth recording, because the same trap will catch the next person:

- The interactor tests use a mock permission checker. They prove the right resource and action are **requested** — never that the policy can grant it.
- `ci-policies` only triggers on changes under `internal/rbac/**`, which this PR did not touch until now, and it only checks that generation succeeds.

So `internal/rbac` now has its own test asserting every action the interactor checks has a rule, with the roles each grants, plus a guard that the document split did not widen `ResourceProject.edit`. Both verified against mutants: removing the `read` rule fails with *"checks read on projectDocument; without a rule it is denied for everyone"*, and removing `writer` from `edit` fails too.

`make gen-policies` runs clean. The generated `policies/` directory is build output and stays untracked.

One open question for you: `read` currently grants **reader** as well, on the basis that anyone who can see a project should see its version history. Say if you would rather that were writer and above.

server/api/internal/rbac/definitions.go
```diff
@@ -23,9 +23,14 @@ const (
 	ResourceParameter     = "parameter"
 	ResourceProject       = "project"
 	ResourceProjectAccess = "projectAccess"
-	ResourceTrigger       = "trigger"
-	ResourceUser          = "user"
-	ResourceWorkspace     = "workspace"
+	// ResourceProjectDocument covers the collaborative document behind a project
+	// (version history, rollback, import, copy). Separate from ResourceProject
+	// because editing the canvas is project CONTENT, which writers do, while
+	// ResourceProject's edit is project settings, deliberately maintainer+.
```
pyshx: Please remove these

server/api/internal/rbac/definitions.go
```diff
@@ -23,9 +23,14 @@ const (
...
+	ResourceProjectDocument = "projectDocument"
```
pyshx, line 26: What's the difference between project, projectAccess and projectDocument? As far as I know there's no projectDocument Resource defined.

server/api/internal/usecase/interactor/common.go
```diff
@@ -57,8 +57,13 @@ func NewContainer(r *repo.Container, g *gateway.Container,
 		Trigger:       NewTrigger(r, g, job, permissionChecker),
 		User:          NewUser(GQLClient.UserRepo),
 		UserFacingLog: NewUserFacingLogInteractor(g.Redis, r.Job, permissionChecker),
-		Websocket:     client,
-		WorkerConfig:  NewWorkerConfig(r, permissionChecker),
+		// The GraphQL surface gets the authorizing wrapper, not the bare client:
+		// every document operation is addressed by project id and must be checked
+		// against that project's workspace. NewProject above keeps the raw client
+		// on purpose, so project deletion (already permission-checked) does not
+		// re-resolve a project that is mid-delete.
```
pyshx: Please remove this.

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,132 @@ package interactor
...
 	"github.com/reearth/reearth-flow/api/internal/usecase/interfaces"
+	"github.com/reearth/reearth-flow/api/internal/usecase/repo"
+	"github.com/reearth/reearth-flow/api/pkg/id"
 	ws "github.com/reearth/reearth-flow/api/pkg/websocket"
+	"github.com/reearth/reearthx/rerror"
 )
 
+// Websocket authorizes the document operations against the owning workspace of
+// the project they address, then delegates to the client. Internal callers stay
+// on the raw client: ProjectDeleter already checks ActionDelete.
```
pyshx: We don't need this, Its self-explanatory.

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,132 @@ package interactor
...
 type Websocket struct {
-	client interfaces.WebsocketClient
+	client            interfaces.WebsocketClient
+	projectRepo       repo.Project
+	permissionChecker gateway.PermissionChecker
 }
 
-func NewWebsocket(client interfaces.WebsocketClient) *Websocket {
-	return &Websocket{client: client}
+// NewWebsocket wraps client with workspace authorization.
```
pyshx: remove

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,132 @@ package interactor
...
+		client:            client,
+		projectRepo:       projectRepo,
+		permissionChecker: permissionChecker,
+	}
 }
 
-func (i *Websocket) GetLatest(ctx context.Context, id string) (*ws.Document, error) {
-	return i.client.GetLatest(ctx, id)
+// authorize checks action against the workspace owning docID's project. Fails
+// closed at every step, so a malformed id cannot skip the check.
```
pyshx: remove

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,132 @@ package interactor
...
+	if err := i.authorize(ctx, docID, rbac.ActionRead); err != nil {
+		return nil, err
+	}
+	return i.client.GetHistoryMetadata(ctx, docID)
 }
 
-func (i *Websocket) FlushToGCS(ctx context.Context, id string) error {
-	return i.client.FlushToGCS(ctx, id)
+// Rollback prunes every update above the target clock, so an unauthorized call
+// would permanently truncate history.
```
pyshx: 1 liner please

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,132 @@ package interactor
...
 
+func (i *Websocket) FlushToGCS(ctx context.Context, docID string) error {
+	if err := i.authorize(ctx, docID, rbac.ActionEdit); err != nil {
+		return err
+	}
+	return i.client.FlushToGCS(ctx, docID)
+}
+
+// CreateSnapshot backs previewSnapshot and does not write despite the name, so
+// ActionRead is correct.
```
pyshx: Remove

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,132 @@ package interactor
...
+	if err := i.authorize(ctx, docID, rbac.ActionRead); err != nil {
+		return nil, err
+	}
 	return i.client.CreateSnapshot(ctx, docID, version, name)
 }
 
-func (i *Websocket) CopyProject(ctx context.Context, id string, source string) error {
-	return i.client.CopyDocument(ctx, id, source)
+// CopyDocument needs edit on the destination AND read on the source, or a caller
+// could pull another workspace's document into a project they own.
```
pyshx: Remove

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,132 @@ package interactor
...
-func (i *Websocket) DeleteDocument(ctx context.Context, id string) error {
-	return i.client.DeleteDocument(ctx, id)
+func (i *Websocket) DeleteDocument(ctx context.Context, docID string) error {
+	if err := i.authorize(ctx, docID, rbac.ActionDelete); err != nil {
+		return err
+	}
+	return i.client.DeleteDocument(ctx, docID)
 }
 
+// Close is connection lifecycle, so it carries no permission check.
```
pyshx: Remove

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -3,58 +3,132 @@ package interactor
...
 func (i *Websocket) Close() error {
 	return i.client.Close()
 }
+
+var _ interfaces.WebsocketClient = (*Websocket)(nil)
```
pyshx: Please move this into the corresponding test file.

bnimit:
## Rebased onto merged #2338, and comments addressed in 28ed8484a

### Answering the question first

> What is the difference between project, projectAccess and projectDocument? As far as I know there is no projectDocument Resource defined.

It is defined, but **only in this PR** — that is why it looks missing. `internal/rbac/definitions.go` adds it as part of this change, so it does not exist on `main`.

The three cover different things:

| Resource | Covers | Why it differs |
|---|---|---|
| `project` | the project entity: create, rename, reconfigure, delete | `edit` is maintainer/owner, because renaming a project is privileged |
| `projectAccess` | public sharing tokens | maintainer/owner only |
| `projectDocument` | the collaborative document: version history, rollback, import, copy | `edit` includes **writer**, because editing the canvas is normal content work |

The reason for a separate resource rather than reusing `project`: `project`'s `edit` deliberately excludes writers. If the document operations used it, writers could not save a version, roll back, import or copy — and writers are exactly the people doing the collaborative editing. Adding writers to `project.edit` instead would have widened project renaming and reconfiguration too, which is a bigger change than this PR should make.

It follows the same shape as `asset`, `parameter` and `node`, which already grant writer/maintainer/owner for content work.

### Rebase

#2338 has merged, and it extended `interfaces.WebsocketClient`, so the authorizing wrapper gains the two new methods: `GetNamedSnapshots` (`ActionRead`) and `SaveNamedSnapshot` (`ActionEdit`). Both are in the permission tests, and leaving either unguarded fails them — verified by mutation, which is what the enumerated lists are for.

### Comments

All applied: trimmed in `definitions.go`, `common.go` and `websocket.go`, and the compile-time interface assertion moved into the test file so the production file carries no declaration that exists only for the type checker.

### Verification

Build, tests, e2e and `golangci-lint` clean using CI's `--new-from-rev` invocation, and `make gen-policies` runs clean. I also checked that **every commit on the branch builds** — a conflict resolution during the rebase had briefly left markers in the working tree, and I folded the fix into the same commit rather than leaving a broken one in history.

pyshx (approved):

PR #2353, opened 2026-08-06
feat(engine): PLATEAU6 data-convert 01-bldg flattened attributes

asrcpq:
# Overview

This PR implements flattened attributes for PLATEAU6 data-convert workflow by updating workflow and related actions.

## Changes

1. `attribute/table_extractor.rs`: New "Attribute Table Extractor" processor - moves values between nested map/list attribute paths via a per-feature-type source/destination path table, with optional type conversion.
2. `feature/reader/citygml2.rs`: Generalize bool `flattenMeasureTypes` (hardcoded to `uom`) into `flattenLeafAttributes: string[]`, collapsing any leaf whose sole XML attribute is in the list.
3. `geometry/splitter.rs`, `plateau6/01-bldg/workflow.yml`: Rename generated attributes `_split_index` -> `__split_idx` and `_output_base` -> `__output_base` to the dunder convention.
4. `testing/plateau-tiles-test/src/tester/output_files.rs`: New tester asserting must-include/must-exclude (or `"rest"` for closed-world) files under flow output; migrate the old `must_not_exist` json_attributes checks to it.
5. `plateau6/01-bldg/workflow.yml`: Wire in the new Attribute Table Extractor, feeding it a flatten table and nesting parsed CityGML attributes under `cityGmlAttributes`.
6. `plateau6/01-bldg/workflow.yml`: Add `__baseCityCode`/`__wardParentCode` attributes and a `CityCodeBoundaryFilter` to drop features belonging to a different parent ward/city.
7. `plateau6/01-bldg/workflow.yml`: Enable `skipUnexposedAttributes` on the Cesium 3D Tiles writer nodes, shrinking tile output.
8. `plateau6/01-bldg/ward/profile.toml`: Add a closed-world output-files check asserting only the 3 expected 3D Tiles zips are produced.

## Attributes Example

See https://github.com/reearth/reearth-flow/blob/ecc1eb202abc6853667249e036dc8e38e670dfef/engine/testing/data/testcases/data-convert/plateau6/01-bldg/osaka-ward/truth/cesium_lod1_truth.json

## Notes

- both JSON `.attributes` and flattened toplevel attributes are produced (follow PLATEAU4 style). We probably want to keep one only in the future.
- Attribute Table Extractor currently takes JSON/inline as input which follows CSV table converter and plateau4/5 convention, another direction is to make it "nested attribute manager" to separately configure each field directly in the action.
- TODO: Attribute Table Extractor input currently is based on some flat key list (follow PLATEAU4 attribute flattener convention), which does not support array and may silently overwrite destination. Should properly support arrays using like JSONPath and maybe report error at collision.


n4to4 (approved):
👍

Issue #2332, opened 2026-08-03
Snapshot rows in the Version panel are read-only: needs previewNamedSnapshot / restoreNamedSnapshot


bnimit:
The Version panel now lists snapshot-backed versions, but the rows are informational only: no preview click-through and no restore. That is a deliberate constraint, not an oversight, and closing it needs a backend change.

### Why they are read-only

Two unrelated ID spaces meet in the panel:

- `NamedSnapshot.id` — a backend-assigned, per-room snapshot counter
- the raw update-log `version` — a CRDT update clock

The existing mutations are built on the second one:

```graphql
previewSnapshot(projectId: ID!, version: Int!, name: String): PreviewSnapshot
rollbackProject(projectId: ID!, version: Int!): ProjectDocument
```

There is no correct client-side mapping between them, and getting it wrong is not a cosmetic failure. `rollbackProject` reaches `PruneAfter`, which deletes every update above the target clock. Feeding it a snapshot id would prune to an unrelated clock and destroy history. During development a handler that did exactly this was caught in review, so the rows were deliberately left inert rather than shipped with a plausible-looking mapping.

### What is needed

Snapshot-addressed equivalents that never touch the update-log clock:

- `previewNamedSnapshot(projectId: ID!, snapshotId: Int!): PreviewSnapshot` — the state is already retrievable server-side via the snapshot store `GetSnapshotState`, so this is mostly plumbing
- a restore path for snapshots. Worth deciding explicitly whether restore should **append** the snapshot state as a new update (non-destructive, preserves history, symmetric with how import works) rather than prune to it. Non-destructive is the safer default and sidesteps the data-loss class above entirely.

### Pointers

- `ui/src/features/Editor/components/OverlayUI/components/Version/components/VersionHistoryList.tsx` — the read-only comment explaining the constraint
- `server/api/gql/document.graphql` — existing preview/rollback surface
- `server/websocket-go/internal/http/adapt.go` — `GetSnapshotState`, `SaveSnapshot`

bnimit:
Related: #2344 gives snapshot ids and update-log versions distinct types.

[...]

PR #2358, opened 2026-08-07
feat(ui, api): snapshot-backed version history with preview and restore

bnimit:
Makes the project Version History panel list real labelled snapshots instead of one entry per persisted CRDT update clock, and lets users preview and restore them.

## What changed

**Server** — a read-only, snapshot-addressed way to fetch document state:

- `Client.GetSnapshotState` over the existing `GET /api/document/{id}/snapshots/{sid}`
- interactor method authorized with `ActionAny`, since it only reads
- `projectNamedSnapshot(projectId, snapshotNumber)` returning `NamedSnapshotState`

**UI** — the panel becomes interactive:

- selecting a row previews that snapshot
- **Restore** returns the project to it, behind a confirmation dialog
- rows show their label (or "Autosaved") plus a `Snapshot N` badge

## Why restore is safe

`rollbackProject` reaches `PruneAfter`, which deletes every update above the number it is given, and a snapshot number is an unrelated per-room counter rather than an update-log clock. Feeding one to the other would rebuild the project at an arbitrary point and destroy history. An earlier revision of this panel wired exactly that and it was caught in review.

So restore does not use `rollbackProject` at all. It reads the snapshot state and applies an **inverse update** to the live document:

- every existing update stays in the log, so the pre-restore state remains reachable
- the change reaches other peers as an ordinary edit over the existing websocket, so no forced resync is needed
- restore captures a snapshot of the current state first, because auto-versioning runs only every 15 minutes and recent work may otherwise have no entry to return to

The Yjs logic lives in `yjsRevert.ts`, shared with the corruption-recovery flow rather than duplicated.

## Two bugs fixed that the old tests missed

**Row badges repeated the row's own date** and the snapshot number was never displayed anywhere. Every production snapshot arrives labelled `auto`, so the panel rendered a column of duplicated dates.

**The tests accommodated it.** Two assertions used `getAllByText(...).length >= 1`, true whether a date rendered once or twice. They now pin exact per-row text.

Both new guards are mutation-tested: reinstating the duplicate badge fails two tests, and wiring restore to `rollbackProject` fails the guard.

## Notes for reviewing

> **The Cloud Run preview will not exercise preview or restore.** `deploy-ui-pr` ships only the UI, and the API it points at does not have this PR's server half yet. Confirmed against it: `Cannot query field "projectNamedSnapshot" on type "Query"`. The snapshot list renders there, but selecting a row and restoring will fail. A local stack with this branch's `server/api` is needed for those two.

Mock mode (`FLOW_MOCK_ENABLED=true`) will not render the app; that repair is tracked separately and deliberately not included here.

Two things that otherwise look like bugs:

- `REEARTH_FLOW_AUTO_VERSION_EVERY` defaults to **15 minutes**, so no new rows appear until it elapses. Override it when testing.
- There is no "save a named version" control in the UI, so every row reads "Autosaved" unless you call the mutation directly:

```bash
curl -s localhost:8080/api/graphql -H 'Content-Type: application/json' -H 'authorization: Bearer <token>' -d '{"query":"mutation{ saveNamedSnapshot(projectId:\"<projectId>\", label:\"before migration\"){ snapshotNumber label } }"}'
```

Entry point is the **⋮** button at the far right of the editor's top toolbar, then **Version History**. The menu is `align="end"` off a button flush against the window edge, so on a narrow window it can open off-screen.

If you saw `"operation denied"` on `projectNamedSnapshots` or `latestProjectSnapshot` in an earlier round, that was #2341 on `main` using an RBAC resource never published to Cerbos. Fixed by #2359, merged and included here.

## Verification

`yarn build`, `tsc`, `yarn lint` (0 errors), `yarn format:check`, `yarn test --run` (46 files / 409 tests) all pass. Server: `go build ./...` clean, 19 packages pass; the 30 failing e2e tests are the pre-existing ones needing `make run-db`.

The interactive flow was driven in a real browser: three rows with correct labels and badges, selecting one changes the header to `Viewing Snapshot: 3` and enables Restore.

## Follow-ups

- #2332 covered snapshot preview and restore and is delivered by this PR
- **Mock server repair** (30 stale operations, missing auth provider, missing `/actions` handler, a fixture keyed `mock-entry-graph` where the canvas expects `main`) — its own PR
- `REEARTH_FLOW_AUTO_VERSION_EVERY` and `REEARTH_FLOW_KEEP_SNAPSHOTS` are undocumented and unset in Terraform, so every environment silently runs the 15m/50 code defaults
- `KeepSnapshots` eviction ignores labels, so named snapshots are evicted too (reearth/ygo#212)
- restore's write travels the collaborative document path, so it is gated by websocket access rather than `ActionEdit`. It grants nothing a reader could not already do by editing the canvas, but this code does not enforce edit permission and the interactor comment says so


bnimit:
## Both reported issues addressed

**The permissions error is not from this PR** — it comes from #2341, already merged. Fixed in **#2359**, which should go in before this one.

Root cause: #2341 introduced a new RBAC resource, `flow:projectDocument`, but Cerbos loads policies from a store this repo never publishes to. `ci-policies` only *compiles* the generated output, and Terraform uploads `.cerbos.yaml` but not the policy files. So the resource existed only in Go code, and Cerbos denies any action on a resource it has never seen.

**Worth checking on your side:** the blast radius is wider than the two queries you hit. `saveSnapshot` is the editor's save action, reached via `useProjectSave` in `features/Editor/hooks.ts`, and it was gated the same way — so **project saving was probably broken too**. One click will confirm, before and after #2359.

**i18n is fixed here** in 2e678a25c. You were right: `yarn i18n` extracts keys but does not translate them, and I only checked that the keys existed rather than that they had values. es is 817/824 and ja 815/822 populated, so an empty value would have shipped raw English into four locales. Translations follow each locale's existing phrasing for these shapes.

## On "has this been tested locally in the UI"

Honestly, no — and I should have said so in the description instead of letting the test counts imply otherwise. I verified types, lint and 283 unit tests, not a running stack. Since the query was denied server-side, the panel could not have rendered for me either.

Once #2359 lands I would rather you exercise it than take my word for it, because a second problem could be hiding behind the first.

ui/src/mocks/schema/typeDefs.ts
```diff
@@ -276,6 +276,13 @@ export const typeDefs = `
     createdAt: DateTime!
   }
 
+  type NamedSnapshot {
+    id: Int!
+    label: String!
+    timestamp: DateTime!
+    size: Int!
+  }
```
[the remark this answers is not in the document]
bnimit: Correct, fixed in fba9c547c. My earlier rename used TypeScript syntax, so it never matched the GraphQL SDL string in this file.

Worth flagging what I found while fixing it: **the mock layer does not build at all.** `Job.userFacingLogsURL` is referenced by a resolver but missing from the mock's `Job` type, which throws at import — so this `snapshotNumber` mismatch would have been the *second* error, not the first. The real schema has that field, so I added it, and the mock layer now imports cleanly (verified with a throwaway probe test).

Nothing in the test suite imports that layer, which is exactly why both mismatches went unnoticed.

ui/src/mocks/schema/resolvers.ts
```diff
@@ -540,6 +540,31 @@ export const resolvers = {
...
+          timestamp: "2024-01-15T09:30:00Z",
+          size: 3072,
+        },
+        {
+          id: 1,
+          label: "initial import",
+          timestamp: "2024-01-01T10:00:00Z",
+          size: 2048,
+        },
+      ];
```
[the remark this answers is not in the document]
bnimit: Correct, fixed in fba9c547c. All three entries now use `snapshotNumber`.

Same root cause as the typeDefs comment: my rename matched TypeScript field syntax, not these object literals.

ui/src/features/Editor/components/OverlayUI/components/Version/recoveryHooks.ts
```diff
@@ -0,0 +1,254 @@
...
+      yDoc.transact(() => {
+        revertUpdate(yDoc, convertedUpdates, getMetadata);
+      });
+      setOpenVersionConfirmationDialog(false);
+      onDialogClose();
+    } catch (error) {
+      console.error("Project Rollback Failed:", error);
+      setOpenVersionConfirmationDialog(false);
+      onDialogClose();
+      return toast({
```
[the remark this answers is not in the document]
bnimit: Confirmed and fixed in fba9c547c, and it was worse than the four lines you flagged.

`handleProjectRollback` sets `isReverting` true, and its `catch` does `return toast(...)` — skipping the reset that sat after the try/catch. Preview is worse: **three** separate paths leave `isLoadingPreview` true (`!previewData`, the missing-workflows guard, and the catch), so a failed preview spins forever.

Both now reset in a `finally` block rather than relying on falling through to the end, which covers the guards and the catch together. This is inherited code, but it is a real user-visible bug and the fix is small.

ui/src/features/Editor/components/OverlayUI/components/Version/RecoveryDialog.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    selectedProjectSnapshotVersion: undefined,
+    previewedProjectSnapshotVersion: undefined,
+    isLoadingPreview: false,
+    isCorruptedVersion: false,
+    openVersionConfirmationDialog: false,
+    setOpenVersionConfirmationDialog: vi.fn(),
+    handleProjectSnapshotSelect: vi.fn(),
+    handleProjectRollback: vi.fn(),
+  }),
+}));
```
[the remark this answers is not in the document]
bnimit: Fair, and fixed in fba9c547c. `previewDocRef` was the important omission: `handleDialogClose` dereferences `.current`, so a missing ref would have thrown rather than failed clearly.

The mock now mirrors all 14 keys the component destructures, and I checked that mechanically rather than by eye:

```
destructured: 14 | missing from mock: NONE
```

ui/src/features/Editor/components/OverlayUI/components/Version/components/VersionHistoryList.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+  test("renders every snapshot, newest first", () => {
+    const { container } = render(<VersionHistoryList snapshots={snapshots} />);
+
+    const labels = Array.from(
+      container.querySelectorAll("p.flex-2.self-center"),
+    ).map((el) => el.textContent);
+    // Newest (id 2, later timestamp) should be rendered before the older one.
+    expect(labels).toHaveLength(2);
+    expect(labels[0]).toBe("before migration");
+  });
```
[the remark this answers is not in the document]
bnimit: Agreed, fixed in fba9c547c. It now compares visible text positions instead of querying `p.flex-2.self-center`.

Verified in both directions, since a test like this needs to be sensitive to the right thing and insensitive to the wrong one:

- reversing the sort comparator → **fails**
- restyling the row's classes → **still passes**

billcookie (approved):

server/api/gql/document.graphql
```diff
@@ -54,4 +64,8 @@ extend type Query {
   projectHistory(projectId: ID!): [ProjectSnapshotMetadata!]!
   projectSnapshot(projectId: ID!, version: Int!): ProjectSnapshot!
   projectNamedSnapshots(projectId: ID!): [NamedSnapshot!]!
+  # Read-only, and the basis for both snapshot preview and restore. Restore is
+  # applied client-side as an inverse update, so it needs no destructive
+  # counterpart to rollbackProject.
```
pyshx: Please remove this

server/api/gql/document.graphql
```diff
@@ -36,6 +36,16 @@ type NamedSnapshot {
   size: FileSize!
 }
 
+# One snapshot's stored state, addressed by snapshotNumber.
+#
+# Deliberately separate from ProjectSnapshot: that type is keyed on the raw
+# update-log clock, which previewSnapshot and rollbackProject consume. The two
+# numbering spaces are unrelated, so they must not share a type.
```
pyshx: Make it a one liner

server/api/internal/adapter/gql/resolver_document.go
```diff
@@ -50,6 +51,24 @@ func (r *queryResolver) ProjectHistory(ctx context.Context, projectId gqlmodel.I
 	return nodes, nil
 }
 
+// ProjectNamedSnapshot reads one snapshot's state by its per-room snapshot
+// number. Never pass a snapshotNumber to previewSnapshot or rollbackProject:
+// those consume the update-log clock, a different numbering space.
```
pyshx: Can you make this one liner.

server/api/internal/infrastructure/websocket/client.go
```diff
@@ -611,3 +611,55 @@ func (c *Client) ImportDocument(ctx context.Context, docID string, data []byte)
...
 	return nil
 }
+
+// GetSnapshotState fetches one snapshot's stored state, addressed by the per-room
+// snapshot number rather than the update-log clock.
+//
+// Read-only by construction. This is what makes snapshot preview and restore safe:
+// unlike Rollback, which reaches PruneAfter and deletes every update above a
+// clock, this only reads. Restore is then applied client-side as an ordinary
+// inverse update, so no history is destroyed and peers converge normally.
```
pyshx: One Liner Please

server/api/internal/infrastructure/websocket/dto.go
```diff
@@ -48,3 +48,12 @@ type snapshotItemResponse struct {
 type saveSnapshotRequest struct {
 	Label string `json:"label"`
 }
+
+// snapshotStateResponse mirrors websocket-go's SnapshotStateResponse. Updates
+// arrives as a JSON int array (the server's UpdateBytes marshaller); encoding/json
+// decodes that into []byte as readily as it does base64.
```
pyshx: One liner

server/api/internal/usecase/interactor/websocket.go
```diff
@@ -89,6 +89,22 @@ func (i *Websocket) SaveNamedSnapshot(ctx context.Context, docID, label string)
...
 
+// GetSnapshotState reads one snapshot's stored state, addressed by the per-room
+// snapshot number. ActionAny, not ActionEdit: this only reads, so a reader may
+// preview a version.
+//
+// Restore is built on this same read. The write half never comes through here:
+// the client applies the snapshot as an ordinary inverse update to the live
+// document, which is why restore preserves history instead of pruning it. That
+// write inherits whatever the collaborative document path enforces, so this
+// method deliberately does not claim to authorize it.
```
pyshx: One liner Please

server/api/internal/usecase/interfaces/common.go
```diff
@@ -18,6 +18,10 @@ var (
 	ErrFileNotIncluded error = errors.New("file not included")
 	ErrFeatureNotFound error = errors.New("feature not found")
 	ErrInvalidOperator error = errors.New("invalid operator")
+	// ErrSnapshotNotFound: retention (KeepSnapshots) evicts snapshots, so a row
+	// listed a moment ago can be gone by the time it is clicked. Distinct from a
+	// server fault so the UI can say "no longer available" instead of "error".
```
pyshx: Remove this

server/api/pkg/websocket/websocket.go
```diff
@@ -26,3 +26,14 @@ type SnapshotMetadata struct {
 	ID        int64
 	Size      int64
 }
+
+// SnapshotState is one snapshot's stored document state.
+//
+// Addressed by SnapshotID, the per-room snapshot counter, NOT the update-log
+// version that Document and History carry. The two are unrelated backend-assigned
+// id spaces; passing one where the other is expected reads or prunes an unrelated
+// point in history.
```
pyshx: One liner

ui/src/features/Editor/components/OverlayUI/components/Version/components/RecoveryVersionHistoryList.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { formatDate } from "@flow/utils";
+
+// Recovery-mode list: used only from the project-corruption error boundary
+// (see ../RecoveryDialog.tsx) to let a user roll back to a working update-
+// log version when the project will not open at all. This is the pre-
+// snapshot-history behaviour, restored unchanged: `history` entries carry
+// the raw CRDT update-log `version`, which is exactly what
+// previewSnapshot/rollbackProject expect, so preview-on-click and Revert
+// are safe here. Do not point this at NamedSnapshot data — see
+// ../hooks.ts and ../recoveryHooks.ts for why that ID space is different.
```
pyshx: Can you make one liner. The code it self-explanatory

ui/src/features/Editor/components/OverlayUI/components/Version/components/VersionHistoryList.tsx
```diff
@@ -1,25 +1,31 @@
...
+  selectedSnapshotNumber,
+  onSnapshotSelect,
 }) => {
   const t = useT();
-  const previousVersions = history?.filter(
-    (version) => version.version !== latestProjectSnapshotVersion?.version,
-  );
+  // Sort by snapshotNumber, not timestamp: it is a monotonic per-room counter, so
+  // it is the authoritative creation order and is unaffected by the zero
+  // timestamp the server returns when one cannot be parsed.
```
pyshx: Don't need this.

ui/src/features/Editor/components/OverlayUI/components/Version/hooks.ts
```diff
@@ -1,245 +1,167 @@
...
+//
+// What must never happen here is passing a snapshotNumber to previewSnapshot or
+// rollbackProject. Those consume the raw update-log clock, an unrelated
+// backend-assigned numbering space, and rollbackProject reaches PruneAfter,
+// which deletes every update above the number it is given. An earlier revision
+// of this panel wired exactly that and it was caught in review.
+//
+// Restore therefore never calls rollbackProject at all. It reads the snapshot
+// state and applies an inverse update to the live document, so history is
+// preserved and peers converge over the normal websocket path. See ./yjsRevert.ts.
```
pyshx: Please make this a one liner.

pyshx (changes requested):

server/api/gql/document.graphql
```diff
@@ -16,10 +16,12 @@ type ProjectDocument implements Node {
 }
 
 # Project Snapshot history vector
+# Set either version (clock-addressed) or snapshotNumber (snapshot-addressed).
```
pyshx: remove this

server/api/gql/document.graphql
```diff
@@ -54,4 +56,5 @@ extend type Query {
   projectHistory(projectId: ID!): [ProjectSnapshotMetadata!]!
   projectSnapshot(projectId: ID!, version: Int!): ProjectSnapshot!
```
pyshx: We can extend this query for supporting `snapshotNumber`

bnimit:
Both addressed in `6cd69c90e`.

- **Removed the comment** above `ProjectSnapshot`.
- **Folded `projectNamedSnapshot` into `projectSnapshot`**, which now takes an optional `version` or `snapshotNumber`:

```graphql
projectSnapshot(projectId: ID!, version: Int, snapshotNumber: Int): ProjectSnapshot!
```

Combined with the earlier type merge, the PR no longer adds any new type or query for this — it extends what already existed.

Two things worth flagging on the merged shape, since neither is enforceable in the schema:

- Exactly one argument applies per call, so the resolver rejects both-or-neither at runtime.
- `version` stays unset on the snapshot path. A snapshot has no update-log clock, and a fabricated value reaching `rollbackProject` prunes real history.

Verified: codegen and build clean, document e2e tests pass, UI 434 tests pass.

server/api/internal/adapter/gql/resolver_document.go
```diff
@@ -20,15 +22,35 @@ func (r *queryResolver) LatestProjectSnapshot(ctx context.Context, projectId gql
 	}, nil
 }
 
-func (r *queryResolver) ProjectSnapshot(ctx context.Context, projectId gqlmodel.ID, version int) (*gqlmodel.ProjectSnapshot, error) {
-	history, err := usecases(ctx).Websocket.GetHistoryByVersion(ctx, string(projectId), version)
+func (r *queryResolver) ProjectSnapshot(ctx context.Context, projectId gqlmodel.ID, version *int, snapshotNumber *int) (*gqlmodel.ProjectSnapshot, error) {
```
pyshx, line 25: Can we throw an error when both snapshotNumber and version has been sent?

pyshx (changes requested):

server/api/internal/adapter/gql/resolver_document.go
bnimit: Yes — it already did, but the check was `(version == nil) == (snapshotNumber == nil)`, which is easy to miss. Your asking is the evidence that it was too terse, so I split it into two explicit checks with distinct messages in `501fdbbb2`:

```go
if version != nil && snapshotNumber != nil {
    return nil, errors.New("pass only one of version or snapshotNumber, not both")
}
if version == nil && snapshotNumber == nil {
    return nil, errors.New("pass either version or snapshotNumber")
}
```

Also added a test for both cases. The schema cannot express "exactly one of these", so these guards *are* that constraint and were previously unasserted. I confirmed the test has teeth by deleting the both-sent guard: it fails, and the call reaches the usecase layer instead of being rejected.

pyshx (approved):

PR #2362, opened 2026-08-17
feat(engine): add new-geometry support to GeometryCoercer

n4to4:
# Overview

Makes the `GeometryCoercer` action work under the `new-geometry` feature, so workflows that coerce geometry between line strings, polygons, and triangular meshes can run on the new geometry world.

## What I've done

- Add a `Coerce` op to the geometry crate covering the new leaf types (faces, surfaces, volumes, triangle meshes, collections)
- Wire the action's `new-geometry` path to it, with the old path left untouched behind `cfg`
- Rename the parameter enum `CoerceTarget` to `CoercionTarget` and regenerate the schemas

## What I haven't done

- No workflow-level integration test; coverage is unit tests in the geometry crate and the action

## How I tested

Unit tests for the op (per-target rings, elevation carry-over, idempotency, per-member frames and attributes) and for the action (pass-through, identity/attributes preserved, one feature in one feature out).

## Memo

- Semantics follow FME's `GeometryCoercer`: geometry the target does not apply to leaves unchanged via `features` instead of erroring, and a multi-part geometry is coerced member by member so it no longer fans out into one feature per part as the old CityGML path did.
- `UnsupportedOperation` is reused as the "nothing changed" signal rather than a failure, which is why the action logs it at debug level.
- `CoercionTarget::TriangularMesh` on a volume keeps it a volume with a triangulated boundary, matching `Triangulate` rather than returning a bare mesh.
- The rename only changes the JSON Schema definition name; the accepted enum values are unchanged, so existing workflows keep working.


shunski (approved):
👍

PR #2366, opened 2026-08-18
fix(engine): keep proj.db reachable when grid search paths are set

n4to4:
# Overview

Coordinate transformation fails on installations where `proj.db` lives under PROJ's own default search path, because the grid search paths introduced in #2346 replace that path instead of adding to it. This restores the database lookup so reprojection works again on such installations.

## What I've done

- Ask a context with PROJ's untouched defaults where it resolved `proj.db`, and append that directory to the grid search paths we hand to every context, so the database — and the grids an installation ships beside it — stay reachable.
- Add a regression test asserting the resolved search path still holds the directory of the `proj.db` PROJ found on its own.
- Loosen the vertical-datum height tolerance from 0.01 m to 0.05 m: with the installation's own grid directory now searched, the France case may pick RAF18 instead of the embedded RAF20, which puts the point 0.01 m higher.

## What I haven't done

- No fallback when PROJ itself finds no database; that case behaves as it did before.

## How I tested

`cargo make test` on macOS with Homebrew PROJ, which fails without this change at every PROJ call and passes with it.

## Memo

`proj_context_set_search_paths` replaces PROJ's search paths rather than prepending to them, so a Homebrew install whose `proj.db` sits in `share/proj` loses it the moment we point the context at the embedded grid directory. CI did not catch this — worth knowing that this code path is only exercised on layouts where the database is not alongside the grids.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4212,7 +4212,7 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "plateau-tiles-test"
-version = "0.2.113"
+version = "0.2.114"
 dependencies = [
  "bytes",
  "glam 0.30.10",
@@ -4751,7 +4751,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-log"
-version = "0.0.489"
+version = "0.0.490"
 dependencies = [
  "futures",
  "once_cell",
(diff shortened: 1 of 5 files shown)


shunski (approved):
👍 
Since the proj now searches for system installed proj first, there would be differences between local envs and remove envs. We could consider using `bundled_proj` feature instead to not rely on system proj at all to avoid behavioral differences between machines, but it has its own weaknesses (such as we cannot use newer EPSG codes), so it will need further discussion.

PR #2368, opened 2026-08-19
feat(engine): port Dissolver to new geometry

n4to4:
# Overview

Makes `Dissolver` work under the new geometry world.

## What I've done

- Accept 2D areal geometry (polygons and meshes) in a single coordinate frame; send non-areal, elevated, 3D, and out-of-frame features to `rejected`, as FME's `Dissolver` rejects invalid area geometry.
- `tolerance` snaps near-coincident vertices together before the merge, following FME's tolerance and the legacy `glue_vertices_closer_than` step.
- `useOneFeature` takes attributes from the group's largest member, which is FME's specification; `groupBy`, `dropAttributes`, and `mergeAttributes` keep the existing semantics.
- Fix the invalid `attributeAccumulationMode: useOne` parameter in the plateau4 `02-tran-rwy-trk-squr-wwy` workflows.

## What I haven't done

- FME's remnants and shared-edge outputs, attribute statistics, list generation, and `Connect Z Mode` remain unimplemented, as before.
- Snapping is vertex to vertex, so a gap between two edges with no vertices facing each other stays open.

## How I tested

Unit tests for the dissolve operation and for the processor, the latter under both the default and the `new-geometry` feature.

- A frame mismatch within a group rejects the feature and warns once, rather than failing the node.
- Legacy behavior is deliberately untouched: it still takes `useOneFeature` attributes from the last member.

## Memo

`Dissolver` is used only in plateau4 `02-tran-rwy-trk-squr-wwy` and quality-check `03-tran`. The misspelled parameter was silently ignored and fell back to the same default, so fixing it changes nothing for the legacy world. Output is a single `Polygon` when the group dissolves to one face and a collection otherwise, where the legacy world always emitted a `MultiPolygon`.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4212,7 +4212,7 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "plateau-tiles-test"
-version = "0.2.114"
+version = "0.2.115"
 dependencies = [
  "bytes",
  "glam 0.30.10",
@@ -4751,7 +4751,7 @@ dependencies = [
 
 [[package]]
 name = "reearth-flow-action-log"
-version = "0.0.490"
+version = "0.0.491"
 dependencies = [
  "futures",
  "once_cell",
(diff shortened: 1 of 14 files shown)


shunski (approved):
👍

PR #2251, opened 2026-07-13
refactor(engine): improve Filter action descriptions, params, ports, and tags

KaWaite:
## Overview

Phase 4 action-audit batch for the **Filter** group (7 actions): applies the [action standard](engine/dev-docs/action-standard.md) fixes from the Filter section of `action-review-findings.md`, and deletes that section.

### Descriptions (all 7 actions)
Rewritten as verb-first, third-person, period-terminated sentences (standard §2):
- **Dimension Filter**: "Routes features to output ports based on the number of geometry dimensions."
- **Feature Filter**: "Routes features to named output ports based on user-defined filter conditions."
- **Feature LOD Filter**: "Filters features by Level of Detail (LOD), emitting each to the matching LOD output port."
- **Feature Type Filter**: "Filters CityGML features by their feature type."
- **Input Router / Output Router**: "Forwards features from the parent workflow into a sub-workflow." / "…from a sub-workflow back to the parent workflow."
- **Spatial Filter**: "Filters candidate features based on their spatial relationship to filter geometry."

### Parameters
- **Feature Filter**: `Condition.expr` / `Condition.outputPort` now have title-case titles and descriptions
- **Feature LOD Filter**: `filterKey` gains a "Filter Key" title and a description explaining group-based max-LOD determination
- **Feature Type Filter**: `targetTypes` gains a "Target Feature Types" title and a description with valid-value examples (`bldg:Building`, `tran:TrafficArea`)
- **Input/Output Router**: `routingPort` gains a title and an accurate description of the parent↔sub-workflow wiring semantics
- **Spatial Filter**: params reordered (required → common → edge-case per §3.5); `mergeFilterAttributes` description shortened to 2 sentences; stale `SpatialFilter Parameters` schema title fixed

### Ports — ⚠️ behavioral change
- **Feature LOD Filter**: `up_to_lod0`…`up_to_lod4` → **`max-lod0`…`max-lod4`** (snake_case violated §4.1). The `max-lodN` naming was chosen over a literal `up-to-lodN` conversion because it better reflects the semantics: each port emits features capped at LOD N, including lower-LOD fallback geometry. All plateau4 workflow fixtures swept.
  - Note: the #2244 UI migrators do not cover port renames — old saved projects using Feature LOD Filter edges may need a migrator follow-up.

### Tags
- Dimension Filter: `["2d", "3d"]` → `["3d", "geometry"]` (`2d` not in vocabulary)
[...]

PR #2249, opened 2026-07-13
refactor(engine): Debug action batch — descriptions, params, ports, tags

KaWaite:
## Overview

Phase 4 improvement batch for the **Debug** action group (5 actions), per [action-standard.md](https://github.com/reearth/reearth-flow/blob/main/engine/dev-docs/action-standard.md). Clears the Debug section of `action-review-findings.md`.

### Echo Processor / Echo Sink
- Descriptions rewritten third-person: "Echoes features to logs and passes them through unchanged." / "Echoes features to logs and discards them."
- New `logging` tag (added to the standard's §6 vocabulary in this PR)

### Noop Processor / Noop Sink
- Descriptions rewritten: "Passes features through unchanged." / "Discards all incoming features."
- Tags intentionally stay empty (`debug` would duplicate the category; nothing else applies)

### Feature Counter
- Description → "Assigns a sequential number to each feature, stored in an attribute and optionally grouped by attribute values."
- `countStart`: `i64` → `u64`, now **optional** with schema default `1` (all existing fixtures pass 0 or 1 explicitly — no behavior change)
- `outputAttribute` title → "Count Attribute"
- Param order → `outputAttribute`, `countStart`, `groupBy` (required → common optional → edge-case, §3.5)
- **Removed dead `rejected` output port** — the processor never emits to it; no fixture connects an edge to it (verified across examples/testcases)
- Tags: `aggregation`, `attribute`
- Kept the name "Feature Counter" (the deferred "Feature Numberer" rename idea was reviewed and dropped — no exact FME name collision, and a rename would need an action-name migration path that doesn't exist)

### Docs / generated
- `action-standard.md`: `logging` added to tag vocabulary
- `action-review-findings.md`: Debug section deleted (file contract)
- Schemas regenerated via `cargo make schema-base && cargo make schema-translated`
- `cargo set-version --bump patch`

## Checks

- `cargo make format` / `cargo make clippy` clean
- `cargo make test-rs` green
- `cargo make test-qc` green (182 passed; one unrelated flake in `plateau4_02_bldg_z_bldg_02_surface_intersection_lod0` — nondeterministic intersection pair ordering — passed on rerun and in a full green rerun)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2373, opened 2026-08-20
feat(engine,api): revise the action standard, expose 23 hidden actions

KaWaite:
Follow-up to #2365. The re-verification pass over the 72 already-audited base actions found the structural classes clean, but surfaced three defects — and tracing *why each survived review* pointed at the standard rather than at the reviews. This fixes the standard.

Documentation only. No action metadata changes, so no schema regeneration.

## One rule was actively wrong

§4.3 listed **"missing attribute"** as a case requiring a `rejected` port. It is normally a no-op that must pass through.

This is not theoretical: following that wording added a `rejected` port to `Geometry Replacer` and broke six `test-qc` cases, because its paired producer `Geometry Extractor` deliberately skips writing the attribute, so the consumer legitimately receives features with nothing to restore. Those features had always passed through and instead vanished into an unwired port.

§4.3 now has a failure-versus-no-op subsection with a decision table, the paired-producer check, and the point that mattered most: **adding a port is a data-loss change**, because a new port is unwired in every existing workflow, and the symptom is a downstream count quietly dropping rather than a test failing at the changed node.

## The missing dimension

Every section of the standard asked whether an action is *described* correctly. None asked whether it *works*. That is precisely why `Horizontal Reprojector` and `Vertical Reprojector` were reviewed clean: their names, descriptions, parameters, ports, categories and tags are all fine, and both return `Err` on every feature in the shipped build while still being listed in `base_actions.go`.

New **§7 Functionality and exposure**:

- **7.1 Does it run?** — an action whose only `process`/`start` is under `#[cfg(not(feature = "new-geometry"))]` hits the trait default at `runtime/runtime/src/node.rs:286,375,382` and fails on every feature while looking complete in source. Deliberate `Err` stubs are removals that have not happened.
- **7.2 Should it be exposed?** — `base_actions.go` is a separate file in a separate language, so it drifts silently. Removing a broken action from it is the cheap, reversible mitigation.
- **7.3 Disabled behaviour must not stay documented** — from outside, a feature that is off is indistinguishable from one that never worked. `#[allow(dead_code)]` on an action path is a review flag, not a fix.

The §8 checklist gains a `runs:` line, checked **before** anything is read for style.

## Gaps that were tribal knowledge

| Change | Why |
|---|---|
| Unused-parameter rule now asks *which geometry world* a param is dead in, and catches forwarding-then-dropped params | A `cfg`-gated action can read a param in one world and ignore it in the other; `actions.json` is generated from the default build |
| §3.4 prohibits a root-level `oneOf` parameter block | `apply_parameter_i18n` patches `properties` and `definitions[*]` only, so such an action ships **untranslatable** text permanently. `Geometry Filter` is in this state today |
| §2 rules out naming commercial products in doc comments | `///` comments compile into `actions.json` and ship to users — they are product copy. Cite open specs/implementations (OGC Simple Features, PostGIS, JTS, GDAL/OGR) instead |
| §"How to use" gains an audience statement | Judge actions as general-purpose tools; low usage is a reason to look harder, not to skip |
| §"How to use" gains a prior-art requirement for **design** | §3.1 already required checking prior art for *names* only |

## A changelog, because the last rule change went unpropagated

The standard had no version or changelog. The ACCURACY-BEFORE-STYLE clause — its most important rule — was added mid-audit in #2280 on 2026-07-23, which means:

| Batch | Merged | Clause existed? |
|---|---|---|
| Debug #2249 | 2026-07-14 | no |
| Merge #2250 | 2026-07-14 | no |
| Filter #2251 | 2026-07-16 | no |
| Output #2279 | 2026-07-23 | no |
| Input #2280 | 2026-07-23 | added it |

Roughly 25 actions were reviewed under a standard missing its central rule, and nothing re-opened them. (The re-verification pass has since covered all 25 and found them structurally clean.)

Changelog entries now state when a change invalidates earlier verdicts. Related: deferred findings must record **what would unblock them** — the reprojector removal was correctly deferred while its replacement was gated out of the shipped build, the gate later lifted, and nothing reopened the item.

## Verification

- Both factual claims added to the standard were checked against the code rather than written from memory: the trait-default error text at `node.rs:286,375,382`, and `apply_parameter_i18n`'s reach in `cli/src/utils.rs`.
- Renumbers the checklist §7 → §8; the only in-repo reference (`action-review-findings.md:7`) is updated. Section references in `boundary_extractor.rs` and `geometry_part_extractor.rs` point at §2/§3.4/§3.5/§4.3 and remain valid.
- Version bumped: any change under `engine/` trips the version check in `ci_engine.yml`.


engine/dev-docs/action-standard.md
```diff
@@ -128,6 +134,12 @@ enum AreaType {
 
 A plain `enum` with no doc comments produces no per-variant descriptions and should be converted to this pattern. A comprehensive property `description` that names and explains all variants is acceptable only when the enum has two or three self-describing values and the description remains one sentence.
 
+**Keep mode enums inside a property — never make the parameter block itself a `oneOf`.** A Rust enum used *as the whole parameter type* (`#[serde(tag = "...")] enum FooParam`) generates a schema whose root is a `oneOf` rather than an object with `properties`. Translation cannot reach the variants: `apply_parameter_i18n` (`cli/src/utils.rs`) patches the root's own title/description, root `properties[*]`, `definitions[*].properties[*]`, and `definitions[*].oneOf|anyOf` variants — and that last traversal is scoped *inside* the `definitions` object, so a `oneOf` sitting at the schema root is never visited.
```
KaWaite, line 137: Verified and fixed in 35ffff53d — thank you, the omission was real.

`apply_parameter_i18n` does iterate `&["oneOf", "anyOf"]` at `cli/src/utils.rs:159`, and the text listed only `oneOf`.

Checking it also surfaced a second imprecision that was mine, and a more consequential one. I had written that an action shaped as a root-level `oneOf` ships "untranslatable titles and descriptions". The block's own title and description **do** translate — the `""` key patches the root node directly at `:123-125`. Only the *variants* are unreachable, because the `oneOf`/`anyOf` traversal is nested inside `if let Some(definitions) = schema.get_mut("definitions")`, so a `oneOf` at the schema root is never visited.

`Geometry Filter` demonstrates it: its Japanese entry has a translated block header (`"Geometry Filter パラメータ"`) and no variant entries at all.

The rationale survives intact and is actually stronger stated correctly — the action looks localised while the three mode labels the user chooses between stay in English permanently. Which is a fitting thing for a review to catch on a PR about descriptions matching code.

PR #2384, opened 2026-08-20
feat(engine): port Elevation Extractor to new geometry

n4to4:
# Overview

Ports `ElevationExtractor` to the `new-geometry` world.

## What I've done

- Add an `Elevation` op to the geometry crate and implement it for every leaf, collection, and solid.
- Wire the action to it behind `#[cfg(feature = "new-geometry")]`, leaving the legacy path untouched.

## What I haven't done

- `Csg` reports no elevation: an unevaluated boolean tree has no boundary of its own.

## How I tested

Unit tests for the op across all geometry variants, plus tests for the action's three outcomes (elevation written, nothing to read, non-finite value).

- The definition of "representative vertex": first vertex in nesting order, matching FME's `ElevationExtractor` rather than min/max z. Meaningful only for geometries lying at one elevation.

## Memo

- Behavior on the unhappy paths is pass-through, not drop: a feature with no readable elevation still leaves by `features` without the attribute, because the graphs this action sits in check the geometry downstream and dropping would silently exempt it. A non-finite z is logged instead of failing the node.
- Meshes read the first *face*'s first vertex, not the vertex pool's head — the pool order is unrelated to traversal order. Solids read the exterior shell only, so voids never win.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4212,7 +4212,7 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "plateau-tiles-test"
-version = "0.2.127"
+version = "0.2.128"
 dependencies = [
  "bytes",
  "glam 0.30.10",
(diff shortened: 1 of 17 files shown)


shunski (approved):
👍

PR #2395, opened 2026-08-21
refactor(engine): audit the geometry extractor actions

KaWaite:
Audits five actions against `engine/dev-docs/action-standard.md` and re-exposes them in the palette: **Hole Counter**, **Hole Extractor**, **Geometry Coercer**, **Offsetter** and **Boundary Extractor**. Palette goes 59 → 64; pending audit drops 20 → 16.

`Boundary Extractor` had never appeared in any audit batch. The other four were the findings file's Group A, whose section this PR deletes per that file's contract.

## The headline

Three of the five already carried standard-compliant metadata, because their new-geometry ports were written to the standard. `Offsetter` needed **nothing** — all three offsets are read and applied, the documented default of zero matches the implementation, and the per-axis units are right. What was left was mostly in what the good prose did **not** say, and in the translations.

## Per action

**`Hole Counter`** — the description was wrong twice beyond style. The action is not limited to polygons, and in the shipped build it now *always* writes the attribute: a point, or a feature with no geometry, records `0` where the legacy build passed it through untouched. Kept as correct (zero is a real answer for a counter, and it makes the output attribute unconditional) and now documented. Zero tags confirmed correct — `Vertex Counter`, `Area Calculator`, `Bounds Extractor` and `Coordinate Extractor` are all untagged siblings doing the same kind of work.

**`Hole Extractor`** — understated itself: the exterior ring *always* leaves too, so this splits each face into its rings rather than extracting only holes.

- **Port `outershell` → `exterior`** (breaking). OGC Simple Features reserves "shell" for solids, which is how this suite already uses it — `Boundary Extractor`'s "the bounding shells of a volume", and `Geometry Validator`'s `shellOrientation`. `outershell` was the only place the word described a polygon ring. `exterior` is the spec term and pairs with the existing `hole` port.
- **The name is kept, deliberately.** A rename to `Ring Extractor` was proposed and rejected on prior art: "ring extractor" is not an established term in PostGIS, GDAL/OGR, GEOS, JTS, shapely or QGIS, whereas "hole" is the established user-facing term and "interior ring" the API term. The description now carries both, as PostGIS glosses them.
- **Its `es`, `ja` and `zh` descriptions all claimed it adds holes as attributes.** It has never written an attribute — it emits them as features. The English was merely vague, so this was introduced in translation, most likely by copying `Hole Counter`'s phrasing. Corrected in all four languages.

**`Geometry Coercer`** — `targetType` traced to all three branches and §4.3 confirmed correct (a geometry the target does not apply to passes through on `features`). Text only: dropped a redundant verb, added the terminating period, and the parameter block no longer restates the action name.

**`Boundary Extractor`** — implementation clean. Removes the 27-line `AUDIT NOTE` left by the Geometry A batch: all four of its leads were resolved by the new-geometry port (#2369), and its closing paragraph asserted that ports "cannot vary by parameter", which is not true — `builder_dag.rs` derives ports from `with` at runtime — so the note was carrying a false constraint as well as a stale one. Its legacy params survive but `parameter_schema` returns `None` under new-geometry, so the shipped schema is honest; left as a migration artifact.

## Drive-by fix worth reading

The Dissolver `area` → `features` rename in #2394 removed the `AREA_PORT` static, orphaning `once_cell::sync::Lazy` in that module. That leftover import is a clippy **error** under `-D warnings` — but only in the legacy feature set, so **no check we run catches it**: `cargo make clippy` runs the workspace with `new-geometry` unified on, and CI's `--no-default-features` `test-rs` pass reports it as a warning rather than a failure. This is the mirror image of the known "never verify a new-geometry action with `-p`" trap — the workspace command hides legacy breakage, and `-p` hides new-geometry breakage. Fixed here in its own commit; the lint gap itself is unfiled.

## Verification

| Check | Result |
|---|---|
| `cargo make format` | clean |
| `cargo make clippy` | pass |
| `cargo make test-rs` | 3882 passed, 0 failed (97 suites) |
| `cargo make test-qc` | 12 passed, 179 ignored, 0 failed |
| `cargo test -p workflow-tests -- --test-threads=4` | **185 passed, 0 failed** |
| `go test ./internal/app/` | pass |
| `schema-base` + `schema-translated` re-run | no-op, generated files in sync |

The port rename touches 11 `fromPort:` edges across the plateau4 quality-check workflows and `graphs/surface_validator.yml`, and **zero** truth/results fixtures. `test-qc` ignores almost all of those workflows (179 of 191), so the legacy `workflow-tests` run is the one that actually executes them — hence it is listed above. CI does not run it.

## Breaking

`Hole Extractor`'s `outershell` output port is now `exterior`. Per the project's standing decision, port renames ship without a migration path; the in-repo fixtures are updated in this PR.


n4to4 (approved):
LGTM 👍

PR #2404, opened 2026-08-24
feat(engine): migrate plateau4 qc bldg to new geometry

shunski:
# Overview
In this PR, plateau 4 QC bldg workflow is rewritten to run under new-geometry.
Some features of the workflow is still unimplemented, and will be implemented after necessary actions are implemented in new-geometry.
All workflow E2E tests are re-wired for the part of the workflow that are migrated.

## Engine-side minor changes
There are a few minor changes to the engine side:
- `Coordinate Frame Reprojector` now takes the target EPSG parameter as an expression, instead of taking it as a `u16`.
- `Geometry Validation` splits the `ShellManifold` validation into two separate validations, namely connectedness and water-tightness. These checks were already implemented before, but each of them is now a separate check to give a richer error granularity for downstream actions. 
- `FeatureCityGMLReader` conditionally retains the attribute of the incoming attributes.
- `BuildingUsageAttributeValidator` is changed to read non-flattened CityGML attributes, in order to work with the new `FeatureCityGML2Reader`.

## What's not implemented
The following features of the workflow is not implemented yet:
- Mesh area check (図郭不正). It needs the line segment reprojection with curvature into account, which is not implemented.
- L-bldg-02 Connectivity check: We need an action to check connectivity, which is yet to be implemented.

## Notes
- Some workflow tests' answers are modified. This is because the new validation actions spit out some additional data that weren't previously spit out. 
- The actual Plateau 4/5 workflow is under `plateau5` branch, and will no longer be retained in the main branch after this PR is merged.


n4to4 (approved):
LGTM 👍

PR #2413, opened 2026-08-25
refactor(engine): audit the Elevation Extractor and Shapefile actions

KaWaite:
Audits `Elevation Extractor`, `Shapefile Reader` and `Shapefile Writer` against
`engine/dev-docs/action-standard.md`. All three are confirmed Migrated in the new-geometry
tracker (FLOW-DEV-182) and run in the shipped build.

`Elevation Extractor` is added to the palette. `Shapefile Reader` and `Shapefile Writer` were
already exposed; this corrects the metadata they were exposed with.

## Elevation Extractor

- The description claimed it extracts "the elevation of a feature's geometry". The value is the
  z of the geometry's first vertex in nesting order, which describes the geometry as a whole only
  when the geometry lies at one elevation. Rewritten to say so, and to state that a geometry
  carrying no elevation passes through without the attribute — behaviour that had no user-facing
  mention.
- `outputAttribute` now documents that it is left unwritten when there is no elevation.
- The Spanish and Chinese descriptions described different behaviour again ("extracts the first z
  coordinate value of an entity"), and French was an untranslated placeholder.
- Added to `base_actions.go`. Its port (#2384) moved it out of the does-not-run bucket without
  changing its exposure, so it has been running and unlisted since.

## Shapefile Reader

- The description named the archive's members without saying that a missing `.shx` is read
  anyway. It now states the three files the format defines — main file, index file and dBASE
  table, all three mandatory per spec — and records the missing-index tolerance as a tolerance.
  An earlier revision of this PR cut `.shx` from the requirement on the grounds that
  `archive.rs` warns and reads shapes in order without one; that was reverted in review, because
  an archive with no `.shx` is not a valid shapefile even though we read it, and describing the
  two-file case as the requirement presents spec-invalid input as normal. Updated in English and
  all four translations.
- The parameter block picks up the pairing rule, previously undocumented: components are matched
  by name, and the first shapefile is read when an archive holds more than one.
- `force2D` documented only that it ignores Z. It also fails the read outright on a multipatch,
  which has no 2D form. Now stated.
- `allowEmptyPath` was titled "Allow Null Path" and described only the null case, while the
  implementation treats an empty string as absent too. Retitled "Allow Empty Path" and reworded.
  Its `alias = "allowEmptyPath"` duplicated the name `rename_all = "camelCase"` already
  generates, and is removed; the wire name is unchanged.
- The parameter block was titled `ShapefileReader Parameters`, against the `<Action Name>
  Parameters` convention every other audited action follows.

`encoding`'s documented case-insensitivity is verified accurate and is struck from the
findings file's "still to verify" list.

## Shapefile Writer

- `output` is documented as a directory, which it is — but the file sets inside it are named
  after the `groupBy` key, and with no grouping that name is the literal `null`. Both `output`
  and `groupBy` now describe this. The naming scheme itself is a behaviour change and is filed
  as its own follow-up rather than done here.
- The description no longer names a vendor (§2), matching the reader's wording.
- The parameter block was titled `ShapefileWriter Parameters`.
- Spanish, French and Chinese descriptions were stale, and `groupBy`'s description was the old
  English text in all four languages.

## Filed, not fixed

Recorded in `action-review-findings.md` with triggers:

- **Shapefile Writer output naming** — its own PR. The writer also has no new-geometry workflow
  coverage: all eight `08-dem` cases are `skipNewGeometry: true`, so any fix must land coverage
  with it.
- **Readers' file path renders last, on 6 of 9** — `schemars` appends `#[serde(flatten)]`
  properties after the struct's own, so `dataset`/`inline` sort below format options. Left alone
  deliberately: one corrected reader among six is worse than six consistent ones.
- **`inline` cannot work on a binary-format reader** — it evaluates to UTF-8 text, and Shapefile
  Reader accepts only a ZIP archive. Same for GeoPackage, glTF and OBJ. Fixing it means splitting
  `FileReaderCommonParam`, so it belongs with the ordering fix.

## Bucket table, reconciled against main

Batch 6 (#2396) and the Bufferer re-check (#2422) both merged while this was open, and each wrote
the same table rows independently — the semantic conflict git cannot see. Recomputed rather than
resolved by arithmetic:

| Bucket | Count | How it was derived |
|---|---|---|
| Exposed and audited | 76 | `grep -c 'true,' base_actions.go`; every name verified to resolve in `actions.json` |
| Does not run in the shipped build | 17 | Carried forward from #2422 |
| Pending audit | 8 | Counted from the list in the findings file (2 + 4 + 2) |
| Flagged for removal | 2 | Unchanged |
| Retired on design grounds | 2 | Unchanged |

Sums to 105. `Elevation Extractor` leaves the pending-audit list, which is this PR's only delta to
the table; #2422's classification of `Area Calculator` as pending (a spot-check is not a review)
is kept as it wrote it.

**One item filed, not fixed: `CSV Reader` currently has no bucket.** Its port merged as #2405 and
`file/csv.rs` now has a `#[cfg(feature = "new-geometry")] async fn start`, so it runs — but it is
still counted in does-not-run, is not in `base_actions.go`, and is not in the pending-audit list.
It was reviewed in the Input batch (#2280), so by #2422's own rule it owes a re-exposure decision
rather than an audit. Recorded in the findings file with that trigger so it cannot expire
silently. 

## Pre-existing on main: `workflow-tests` fails 37

Not caused by this PR, and worth flagging because **CI does not run this suite**.

36 failures are `Action not found: Coordinate Frame Reprojector` across
`quality-check/plateau4/02-bldg`; CFR's module and factory registration are both
`#[cfg(feature = "new-geometry")]`, so it does not exist in the legacy build, and #2404 rewrote
those fixtures to use it. The 37th is a WKT parse failure in `test_executor_csv_wkt_roundtrip_3d`,
from #2405.

Verified by checking out `origin/main` (`d5e0e788e`) and running one of the failing tests there:
it fails identically. If legacy is now considered dead, those fixtures want `skipNewGeometry`
rather than being left red.

## Checks

`cargo make format`, `clippy`, `test-rs`, `test-qc`, plus `cargo test -p workflow-tests --
--test-threads=4` and `go test ./internal/app/`.


shunski (approved):
👍

Issue #2412, opened 2026-08-25
websocket-go cutover: readiness checklist and recovery tracking (data loss incident 2026-08-24)


bnimit:
Tracking issue for the websocket-go cutover on 2026-08-24 and everything that has to be true before we point production traffic at it again.

## What happened

The Go websocket server cannot fully decode `doc_v2` documents written by the Rust server. `crdt.ApplyUpdateV2` returns `nil`, integrates roughly 55% of the content, and parks the rest in the pending queue permanently. The server then treats that partial document as the room state and flushes it back over the original object in GCS.

Measured against the reference implementation on 94 real production documents:

| decoder | nodes | edges | documents wrong |
|---|---|---|---|
| yjs 13.6.32 | 974 | 951 | 0 |
| ygo v1.41.0 (deployed) | 535 | 451 | **63 of 94** |
| ygo v1.49.0 (latest) | 535 | 451 | 63 of 94 |

The trigger is GC structs, which appear once a document has real deletion history. Upstream: reearth/ygo#231.

## Damage

| | projects with a document | rewritten by websocket-go | pre-cutover copy recoverable |
|---|---|---|---|
| prod | 809 | 15 | 15 of 15 |
| dev | 862 | 7 | 5 of 5 (2 are new projects) |

Clear content loss in 8 of the 22. The remaining ~1,649 projects still hold their original bytes only because nobody has opened them since the cutover. **Each one is damaged the first time it is opened while websocket-go serves traffic.**

Recovery deadline: GCS soft-delete retention is 7 days, so hard delete begins **2026-08-31**.

## Immediate

- [ ] Repoint the flow websocket url_map back at `reearth-flow-websocket` (Rust) in dev and prod. Reverts the one-line `default_service` change from eukarya-inc/infrastructure#2854.
- [ ] Recover the 22 affected projects from soft-deleted generations, merging the restored state with the current live state so post-cutover work is preserved. Deadline 2026-08-31.

### Ours

- [ ] **Refuse to persist a partial decode.** `internal/gcs` should fail the load when `doc.PendingStats().Items > 0` after decoding a `doc_v2`, rather than serving a truncated document that the next flush writes back. A stored `doc_v2` is a complete self-contained snapshot, so leftovers can only mean the decoder lost something. Branch `fix/ws-refuse-partial-decode` is ready with tests. This is the single change that turns this class of failure from data loss into a visible outage, and it is worth landing regardless of the upstream fix.
[...]

PR #2416, opened 2026-08-25
fix(websocket-go): refuse to serve a partially-decoded doc_v2

bnimit:
Stacked on #2415 — **merge that first.** Refs #2412.

## Why this exists

The 2026-08-24 cutover destroyed data through two independent failures. One was a decoder bug (reearth/ygo#231, fixed in v1.49.1, picked up in #2415). The other is ours:

**we trusted a `nil` return from `ApplyUpdateV2` and persisted whatever came back.** A half-decoded document became the room state, and the next flush wrote it over the intact original in GCS. 22 projects lost content that way.

Fixing the decoder stops this instance. It does not stop the next one. This does.

## What it does

A `doc_v2` object is a complete, self-contained snapshot written by `EncodeStateAsUpdateV2`, which serialises only the integrated store. A correct decode therefore always leaves the pending queue empty. Items still parked in it can only mean the decoder failed to integrate part of the document, and what we're holding is smaller than the bytes on disk.

All four V2 decode sites in `internal/gcs` now go through one helper:

```go
func docFromV2(v2 []byte) (*crdt.Doc, error) {
	doc := crdt.New()
	if err := crdt.ApplyUpdateV2(doc, v2, nil); err != nil {
		return nil, err
	}
	if ps := doc.PendingStats(); ps.Items > 0 {
		return nil, fmt.Errorf("%w: %d items left unintegrated, awaiting %d clients; refusing to serve a truncated document",
			ErrPartialV2Decode, ps.Items, len(ps.MissingFor))
	}
	return doc, nil
}
```

The room refuses to open, loudly, and the stored bytes stay untouched. That is the safe half of the trade: the alternative is silently overwriting someone's work.

## Scope, deliberately narrow

Only the **full-state base decode**. The incremental update objects folded on top of that base are not covered, because pending items are legitimate there — one update can reference structs that another update carries. Applying the same check to the fold would produce false refusals.

## Cost today: zero

| ygo version | production documents | refused by the guard |
|---|---|---|
| v1.41.0 (what shipped) | 109 | **73** |
| v1.49.1 (#2415) | 109 | **0** |

So on top of #2415 this is a no-op, which is exactly what a backstop should be. Merging it **without** #2415 would turn the silent data loss into a refusal-to-open for most projects, hence the stacking.

## Tests

Three, and the main one is written as an **invariant** rather than a snapshot of today's behaviour: `docFromV2` either returns the whole document or an error, never a quietly smaller one. That holds before and after the ygo fix, so it does not need rewriting when the decoder changes.

```
--- PASS: TestDocFromV2_NeverReturnsATruncatedDoc
--- PASS: TestDocFromV2_AcceptsACompleteDecode
--- SKIP: TestDocFromV2_GuardCatchesWhatApplyUpdateV2Hides
    ApplyUpdateV2 now decodes this fixture completely; ygo#231 appears fixed
```

The third self-skips on a fixed decoder and is kept as a regression canary: if a future decoder change reintroduces silent truncation, it runs again and asserts the guard catches it. A skip there is the healthy state.

The fixture is a GC-bearing V2 update generated by `yjs@13.6.32` with **no production data in it**, embedded as hex.

## Verification

- `GOWORK=off go test ./...` green across all ten packages
- `gofmt` clean
- Also worth noting for the reviewer: `ErrPartialV2Decode` is exported so callers can distinguish it from a transport error; the failure surfaces through `Load`'s existing `gcs load failed` ERROR log with the room id, so affected rooms are enumerable from logs rather than guesswork.


bnimit:
Rebased onto main now that #2415 is in, so this is down to the guard alone: four files in \`internal/gcs\`, +247/-8, no \`go.mod\` or \`go.sum\`.

The stacking note at the top of the description is now historical — #2415 merged as \`0eff84f40\`, so there is no ordering constraint left on this one.

Re-verified on the rebased branch: `GOWORK=off go test ./...` green across all ten packages, gofmt clean. With v1.49.1 underneath, the guard is a no-op on all 109 production documents (it would have refused 73 of them on v1.41.0), and `TestDocFromV2_GuardCatchesWhatApplyUpdateV2Hides` self-skips with `ygo#231 appears fixed`, which is the healthy state.

pyshx (approved):

PR #2421, opened 2026-08-27
feat(engine): run plateau4 qc(tran) on new geometry

n4to4:
# Overview

Migrates the plateau4 `03-tran` quality check to `new-geometry`, lifting `skipNewGeometry` from all six of its testcases.

## What I've done

- Rebuilt the surface-check part of the workflow on new-geometry actions, and filled the gaps this exposed in the engine.

## What I haven't done

- `L-TRAN-02` stays `skip: true` (unrelated to this migration).

## How I tested

Test inputs and expected CSV/JSON are unchanged — the six `03-tran` testcases now run in the new-geometry world against the same expectations. Added unit tests for the two engine-side changes.

## Memo

`extractTags` is deliberately left empty: hoisting traffic areas into their own features would break the per-road grouping the overlap check depends on, so each geometry instead carries the object it belongs to as a per-member attribute.

The tolerant path of `check_duplicate_points` used to abort on real input — a reprojected road surface puts hundreds of vertices on one axis value, overflowing a k-d tree bucket — so it is now indexed by grid cell.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -4212,7 +4212,7 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "plateau-tiles-test"
-version = "0.2.146"
+version = "0.2.147"
 dependencies = [
  "bytes",
  "glam 0.30.10",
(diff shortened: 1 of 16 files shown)


shunski (approved):
👍

PR #2426, opened 2026-08-27
fix(engine): give the mapper actions' value parameters a renderable type

KaWaite:
`Attribute Range Mapper` and `Null Attribute Mapper` declared four parameters as `serde_json::Value`. That type means "any JSON value", so schemars emits a leaf with a title, a description, and **no type** — leaving the UI no widget to render. Two of the four (`rangeTable[].outputValue`, `mappings[].replacement`) are required, so they were unfillable required fields.

A sweep of all 166 actions found these two were the only cases. Both had already passed an audit.

## The two actions get different shapes

**`Attribute Range Mapper` — untagged.** `MappedValue` generates an `anyOf` of three typed, titled scalar branches:

```json
"anyOf": [ {"title":"Text","type":"string"},
           {"title":"Number","type":"number"},
           {"title":"True or False","type":"boolean"} ]
```

The written form is unchanged — `outputValue: "#f7f5a9"` still works — and the type written is the type stored, so `"3"` stays text and `3` stays a number.

**`Null Attribute Mapper` — tagged.** `NullReplacement` must offer `Remove` beside the three value types. An untagged version with a unit `Remove` variant (which serde represents as `null`) was rejected: the UI's `simplifyAnyOf` strips `type: "null"` branches out of an `anyOf`, so `Remove` would have been unreachable in the editor.

## Behaviour fixes

- **`Option<Value>` carried opposite meanings for `None`** in `Null Attribute Mapper`: on a mapping it *removed* the attribute, on `defaultReplacement` it *left it unchanged*. Same type, inverted semantics, neither visible in the schema. `replacement` is now **required**, so removing an attribute is stated (`{type: remove}`) rather than implied by omitting a field. `defaultReplacement` stays optional, where omitting it means "leave unchanged" as before.
- Closing the types removes three near-identical `serde_json::from_value` blocks whose `warn!` fallbacks silently dropped the output attribute. Those runtime failure paths existed only because the type was open.

## Tests

`Attribute Range Mapper`'s tests called a helper that **reimplemented** the classification instead of the processor's own code, and the copy had **drifted** — it omitted the boolean coercion `process` performs, so the documented "booleans count as 1 and 0" behaviour was never exercised. The classification is now a single `mapped_value` method that both `process` and the tests call, with added tests for the boolean and numeric cases.

## Translations

The three strings the i18n machinery can reach are translated: `outputValue`'s description, `defaultReplacement`'s, and `replacement`'s — that last one mattered, because the existing Japanese said a null value removes the attribute, which this change makes false. `es`/`fr`/`zh` for those fields were English placeholders and are refreshed.

**The remaining 24 changed strings cannot currently be translated, in any language.** They are the type labels inside `MappedValue.anyOf[*]` and `NullReplacement.oneOf[*]` ("Text", "Number", "True or False", "Remove"), their descriptions, and the two definitions' own titles. `apply_parameter_i18n` matches an enum variant only when the variant carries a literal `enum` key — a tagged object variant nests its tag under `properties.type`, and an untagged scalar branch has `type` but no `enum` — and its `def_i18n` path walks only `definitions[X].properties`, so a definition's own title is unreachable regardless.

This is pre-existing rather than introduced here: `Coordinate Frame Reprojector`'s `BasePoint`, `Coordinate Extractor`'s mode, and five of `HTTP Caller`'s parameters have no i18n entry in any language for the same reason. The trade this PR makes is "unrenderable" for "renderable but English-only in the type picker", which is the better of the two.

The fix is small and symmetric — extend the variant-key derivation in `scaffold_i18n.rs` and `utils.rs` to fall back to `properties.type.enum[0]` and then to `type` — and is deliberately left to its own PR, because it adds i18n keys for every action using these idioms and would collide with HTTP Caller work in flight.

## Also recorded in the findings file

- The CI ratchet that would have caught this: fail when a parameter leaf declares neither `type` nor `enum` nor `const`.
- A **systemic i18n gap, pre-existing and not addressed here**: a tagged enum with *object* variants cannot be translated at all. `apply_parameter_i18n` (`cli/src/utils.rs:157-173`) matches a variant by a top-level `enum` key, which a tagged object variant doesn't have, and its `def_i18n` path only walks `definitions[X].properties`, which a `oneOf` definition lacks. `Coordinate Frame Reprojector`'s `BasePoint`, `Coordinate Extractor`'s mode, and five of `HTTP Caller`'s parameters have no i18n entry in any language as a result. Fixable, and wants its own PR since it changes the i18n skeleton for every action using the idiom.

## Breaking

`Null Attribute Mapper`'s `replacement` is now required and takes `{type: text|number|boolean|remove, value}`. `Attribute Range Mapper`'s written form is unchanged. Neither action is used in any workflow; the one in-repo fixture (`runtime/tests/fixture/workflow/attribute/range_mapper.yaml`) needed no migration.

## Verification

`cargo make format` · `clippy` · `clippy -p reearth-flow-action-processor` (legacy feature set) · `test-rs` (4069 passed, 0 failed) · `test-qc` (44 passed, 0 failed) — all pass. `schema-base` + `schema-translated` re-run clean and idempotent. Zero untyped parameter leaves remain across all 166 actions.


shunski (approved):
👍

PR #2428, opened 2026-08-27
refactor(engine): audit Area Calculator and expose it in the base palette

KaWaite:
## What

Runs the §8 review checklist against `Area Calculator`'s new-geometry implementation, which #2385 landed on 2026-08-27. The action had never had an engine-side pass of any kind — the only prior note on it was a spot-check of `areaType`, a parameter #2385 deleted.

**No correctness defect.** `Area Calculator` is added to `server/api/internal/app/base_actions.go`.

## The implementation, traced

`Area::surface_area` measures every leaf as documented: `Polygon3D` and the 3D meshes take half the magnitude of each ring's Newell vector with holes subtracted and floored at zero; `Polygon2D` and the 2D meshes take their planar area, which for a flat surface is its surface area; `Solid` sums every boundary face, so a void's faces add surface where a polygon's holes subtract it; points, curves and point clouds answer `0.0` rather than refusing. An unevaluated CSG tree is the one geometry the model cannot measure.

`area_report`'s traversals are exhaustive matches with no wildcard arm, so a geometry variant added later is a compile error rather than a silent omission.

The single-`features`-port design is correct rather than a §4.3 gap: the action writes zero for a geometry it cannot measure instead of failing the feature, which is the established model — PostGIS `ST_Area` returns 0 for non-areal geometry, and JTS `Geometry.getArea()` returns `0.0` for every non-areal type.

## Parameter surface

Two properties were describing an implementation that no longer exists, or the wrong thing entirely:

- The **parameter block description** still read "Configure how the area … is measured and stored". Nothing configures how it is measured any more; `areaType` went with #2385. Now "Where the measured surface area is stored on each feature.", matching the shape of its audited sibling `Hole Counter`.
- **`outputAttribute`** spent two of its three sentences restating the action description's void semantics, and never mentioned the behaviour that is genuinely the parameter's own: the attribute is **always** written, recording `0` when the geometry has no area or could not be measured. That promise is asserted in three tests and argued for in code comments, and appeared nowhere a user could see it — so a user reading `area: 0` could not tell a degenerate polygon from an unmeasurable geometry.

Name, description, category and tags are clean. Zero tags is correct and consistent with `Vertex Counter`, `Hole Counter` and `Bounds Extractor`; `3d` would be wrong for an action that measures 2D geometry equally.

## §2 now covers diagnostic registry text

`Area Calculator` is the first audited action to emit structured diagnostics. #2385 wired it to five `geometry.area_*` codes whose `message` and `help` strings live in `schema/error-codes/geometry.toml`. That text reaches users the way a doc comment does — through the diagnostic pipeline and the GraphQL API rather than through `actions.json` — so it is product copy, and the standard had never said so. §2 now holds it to the same accuracy rule and the same ban on naming commercial products, and records the registry's own style conventions so the rules that do not transfer are not applied by mistake.

Reading the five under the new rule found one genuine error, of exactly the kind the new clause predicts — **a string copied from a neighbouring code**:

`geometry.area_not_measurable`'s help said "check that it is a supported, non-degenerate type". Both halves name causes that cannot occur. Degeneracy is never the cause: a degenerate ring measures `0.0` through the success path, and a point or curve answers `0.0` too. And every type is supported bar one. The phrase came from `geometry.footprint_unavailable`, where degeneracy genuinely is a cause. Both this code and `geometry.area_skipped_parts` now name the real cause and point at `CSG Evaluator`, which is what resolves it. The other three were accurate and are unchanged.

Checked and **not** changed: two of the five name `Area Calculator` inside their own `help`, duplicating the diagnostic's `actionType` field. `raster.texture_assignment_failed` already names `Image Rasterizer`, so this is the registry's established convention, and §2 now records it as one.

## Known gap this PR cannot close

**This text has no i18n path.** `schema/i18n/` covers `actions` only, so all 27 error codes ship English in every language. Filed in the findings file with a trigger. It is about to become visible: PR #2401 renders `message` and `help` raw in a diagnostics table whose every other column is translated through the frontend's own i18n, and #2427 has just fixed the wiring that was leaving job diagnostics empty.

## i18n

The worst state seen in any batch. The action description was the **English placeholder in all four languages**; es/fr/zh were English throughout, including the parameter block's own title and description, which are normally the two that do get translated; and `ja` was in the state easiest to misread as finished — block title and `outputAttribute.title` translated, both descriptions English.

All four are now fully translated. ⚠️ **This makes `Area Calculator` the first action with a fully translated es/fr/zh parameter block** — every other audited action leaves those as English placeholders and translates only the action-level description. Worth a reviewer's eye on the language quality rather than just the coverage.

## Also

- Two fixture comments in the solar-radiation workflows quoted the removed parameter ("the planar or sloped area"). The generated `solar-radiation.yml` is unchanged — `yaml-include` strips comments — and `check-generate-examples-cms-workflow` passes.
- The findings file's `areaType` spot-check is deleted. It described a parameter users can no longer set while reading as coverage, and it is the second instance of the trap §7.2 records: a note written before a port describes the implementation the port replaced.

## Bucket table — recounted after merging main

**78 exposed / 6 pending audit**, counted with `grep -c 'true,' base_actions.go`.

This PR originally predicted 77 / 7 on the assumption that #2413 was the only other batch in
flight. That was wrong in the direction it warned about: **two** landed while this was open —
#2413 exposed `Elevation Extractor` and #2425 exposed `HTTP Caller`. On merging, `base_actions.go`
auto-merged correctly to 78 while the bucket table's *text* auto-merged to this branch's own
76 / 8, taken over main's without a conflict being reported. Recounted from the file, and the
addendum now records the sharpened rule: predicting the post-merge number just adds one more
piece of text that merges cleanly and is wrong.

Remaining pending audit (6): `Spatial Filter` · `Feature Reader` · `Feature Writer` ·
`JSON Fragmenter` · `Geometry Filter` · `Attribute Duplicate Filter`. Buckets sum to 105.

## Verification

Baselines taken on this branch's own base before any change, not reused from a previous session — the numbers move as ports merge.

Run before the merge of main, against this branch's own base:

| Check | Baseline | After |
|---|---|---|
| `cargo make test-rs` | 4065 passed / 0 failed / 5 ignored | identical |
| `cargo make test-qc` | 44 passed / 0 failed / 148 ignored | identical |
| `cargo make clippy` (workspace) | 0 warnings | 0 warnings |
| `cargo clippy -p reearth-flow-action-processor --all-targets -- -D warnings` | 0 warnings | 0 warnings |
| `go test ./internal/app/` | ok | ok |

Re-run after merging main (7 commits, including the HTTP Caller rewrite): `format` a no-op,
clippy 0 warnings both ways, `check-schema` and `check-generate-examples-cms-workflow` clean,
`gofmt -l` empty, `go test ./internal/app/` ok. **The `schema-base` + `schema-translated` re-run
after resolving the merge was a perfect no-op**, which is what proves both branches' schema
changes survived it. Test suites are left to this PR's CI.

`workflow-tests` was **not** run. It is broken on `main` independently of this branch (36 `Action not found: Coordinate Frame Reprojector` failures plus one WKT parse failure), and this PR changes no execution path — doc comments, registry help strings, YAML comments and one Go map entry. `Area Calculator` has no case in that suite in any event.

## One thing recorded rather than fixed

`Area Calculator` has **no automated workflow coverage at all**. Its only fixtures are the solar-radiation example workflows, and `testing/data/testcases/` contains no case naming it, so a green `test-qc` says nothing about this action. Its unit tests are thorough by way of compensation, including all five diagnostic paths. Noted in the findings file; not filed as a task.

Separately, verified rather than trusted: `geometry.toml`'s claim that `default_disposition` is inert for these five codes is correct — `ExecutorContext::warn` records unconditionally and only `report()` consults the resolved disposition. The consequence the comment did not draw out is that a workflow-level `ErrorPolicy` override selecting these codes is silently ignored too, which is true of every `warn()` call site rather than of this action. Added to the comment; not changed.


n4to4 (approved):
LGTM 👍

PR #2462, opened 2026-09-09
ci(engine): move the Cargo.toml version check to the end of the lint job

n4to4:
## Overview

Reorder the engine lint job so the version-bump check runs last, after formatting, clippy, and the repo-specific checks.

## What I've done

- Move the `Cargo.toml` version comparison step to the end of the lint job
- Group the remaining steps by purpose (setup / format / lint / repo-specific checks) with comments

## What I haven't done

- No change to what any step actually runs

## How I tested

CI on this branch.

## Memo

The version check fails on any PR that forgets a version bump, which previously masked real lint and format failures. Putting it last means the actual code problems surface first.


shunski (approved):
👍

PR #2463, opened 2026-09-09
fix(engine): nested attribute type should align attribute type

shunski:
# Overview
The attributes use `IndexMap<Attribute, AttributeValue>` though a nested attribute uses `HashMap<String, AttributeValue>`. This PR fixes the nested attribute type to be `IndexMap<Attribute, AttributeValue>`.


n4to4 (approved):
👍

PR #2469, opened 2026-09-11
fix(engine): align plateau6 03-tran QC outputs with original workspace

n4to4:
# Overview

Aligns the plateau6 `03-tran` quality-check outputs with the original workspace.

## What I've done

- Name every result file by package prefix (`03-1_道路`, `03-2_鉄道`, …) via a `packagePrefixes` variable
- Rework 検査結果一覧 to a file/feature-type/LOD row with an instance count
- Split the surface error CSV into エラー数 / エラー内容 and rename the error GeoJSON fields to match
- Stop aborting the run when the objectlist marks nothing as 作成対象 for a package

## What I haven't done

- No test data for the remaining packages (`squr`, `wwy`)

## How I tested

`cargo make test-qc` for the 03-tran cases; added `rwy` and `trk` test cases.

- The summary row granularity (file/feature type/LOD) and how インスタンス数 is counted

## Memo

- Row granularity and column names follow the original workspace, not the plateau4 port.
- One city object can carry several geometries at the same LOD, so instances are de-duplicated by `gml:id` before counting.


shunski (approved):
👍

PR #2471, opened 2026-09-11
refactor(engine): audit Feature Writer and Geometry Filter

KaWaite:
Audits the two migrated members of Group E and exposes both. Rebased onto `main` after #2470 merged.

Both actions had their entire parameter block as a root-level `oneOf`, which §3.4 prohibits because translation cannot reach the variants. Restructuring them turned up that **the fix §3.4 prescribes did not, on its own, work** — see below.

## Geometry Filter

- **Parameters.** `filterType` is now a plain property holding a field-less enum. The YAML is unchanged (`filterType: none` still), so nothing migrated; the schema shape and the i18n reach are what changed. All three mode labels now translate, and the Japanese is in this PR.
- **Description.** Was `"Filter Features by Geometry Type"` — Title Case imperative, no period, and wrong for one of its three modes: `none` filters on *presence*, not type.
- **Port.** `none` → `no-geometry`. The old name read as "nothing" rather than "the features with no geometry". No fixture wired it.
- **Category.** `Geometry` → `Filter`. §5 gives Filter "conditional routing, spatial and type filtering", which is this action exactly, and it matches the audited siblings (`Spatial Filter`, `Dimension Filter`, `Feature Type Filter`). Tagged `geometry`, which only earns its place because the action left the Geometry category.
- Two vendor-name leaks removed from comments. Neither compiled into `actions.json`, so this was not shipping text.

Implementation is clean: all three modes trace to a branch, and every one of the 20 declared ports is reachable from some mode.

## Feature Writer

- **§7.3 — the `citygml` format could never run.** Its arm returns `Err` at build time under new geometry, so a workflow naming it failed before touching a feature, while the palette advertised it plus `lodFilter`/`epsgCode`/`prettyPrint`, all dead in the shipped build. The variant is now behind `cfg(not(new-geometry))`: gone from `actions.json`, still working in the legacy build where `test-conv`'s citygml-roundtrip case covers it. **Trigger to restore: the CityGML Writer sink's port landing (#2355)**, which is what the arm delegates to.
- **Parameters.** `format` is now a property holding a `#[serde(tag = "type")]` enum, so the JSON-only `converter` rides inside the `json` variant and cannot be set against CSV. 461 fixture nodes migrated.
- **Description.** Was `"Writes features from various formats"` — writes *to*, not from, and silent on the two things that actually surprise a user. The output path is a Code expression evaluated **per feature**, so it is the grouping key: one file per distinct result, with no `groupBy`. And the input features are **consumed** — what leaves on `features` is one freshly-built summary per file carrying `filePath` and `rowCount`. Both are now stated.
- **Tags.** `csv` and `json`, matching every other audited writer. Category left `Feature`; cross-cutting finding 3 defers the Feature-vs-Output question for this whole family.
- **Not a delegation to the writer sinks**, which the surface nowhere says. The `csv` and `json` arms are private implementations that have already diverged from the sinks of the same name — the CSV Writer sink emits geometry columns and accounts for failed rows, while this one drops geometry and errors on the first missing field.

### A bug the new tests caught

Dropping the variant's explicit `#[serde(rename = "citygml")]` during the restructure let `rename_all` produce `cityGml`, which would have rejected **every existing legacy CityGML workflow**, including the `test-conv` roundtrip case. The rename is back and the case is covered.

## The i18n tooling fix

Moving a mode enum into `definitions` is necessary but was **not sufficient**. Both halves of the i18n pass identified a variant by a top-level `enum` key on the variant object, which `schemars` emits only for a variant carrying **no fields**. Every variant with sub-parameters — the entire reason to reach for the idiom — was invisible to the scaffold and the applier alike.

What makes this one hard to notice is that the fix and the defect look identical from the schema. `Coordinate Frame Reprojector` was restructured into this shape during its own audit and recorded as resolved, while its `crs` variant and all three `BasePoint` variants shipped untranslated — `BasePoint` had no entry in `ja.json` at all. Nothing failed; the keys simply never existed.

`enum_variant_key` now recognises all three shapes `schemars` produces, and `PropertyI18n` gained a nested `properties` map so a variant's own fields translate too. Re-scaffolding surfaces **88 previously unreachable strings across 13 actions**:

| Action | New keys | Exposed |
|---|---|---|
| HTTP Caller | 34 | yes |
| Rotator 3D | 12 | no |
| Null Attribute Mapper · Coordinate Extractor | 7 each | mixed |
| Coordinate Frame Reprojector | 6 | yes |
| Footprint Replacer | 5 | yes |
| Feature Writer · Geometry Filter | 4 / 3 | yes (this PR) |
| CSV Reader · CSV Writer · Date Time Converter · Geometry Validator · Image Rasterizer | 2 each | mostly yes |

Only Feature Writer's and Geometry Filter's are translated here. The other 86 are seeded English — which is what they effectively were before — and recorded as owing a Japanese pass, HTTP Caller first.

§3.4 and the Changelog are corrected accordingly.

## Also recorded

- **Finding 10 (new):** no action in the workspace sets `deny_unknown_fields`, so a leftover key from a shape change is ignored rather than reported. Deliberately not fixed for one action out of ~100; pinned by a test so the behaviour is at least visible.
- **Deferred, with triggers:** Feature Writer's CSV header comes from the first feature, so a later feature missing a key fails the whole run and extra attributes are dropped — documented, not fixed, since fixing it rewrites committed CSV truth fixtures. Geometry Filter's 20 ports want per-variant declaration, which needs the port-handling revision the engine authors already called for.
- **A correction:** the old Group E note claimed per-variant ports would retire the action-name-keyed special cases in `builder_dag.rs`. Those belong to Output Router and Feature Filter — Geometry Filter is not among them, so the work would *add* a third, not retire two.

## Checks

`cargo make format`, `clippy` (both feature sets), `test-rs`, `test-qc` (153 passed, 0 failed), schema regeneration idempotent, `gofmt` + `go test ./internal/app/`. The legacy citygml-roundtrip `test-conv` case was run directly and passes.


n4to4 (approved):
👍

PR #2474, opened 2026-09-14
feat(engine): add the plateau6 04-frn-veg quality check

n4to4:
# Overview

Adds the city furniture / vegetation (`04-frn-veg`) quality check to the plateau6 (CityGML 3.0 + i-UR 4.0) workflow set, so `frn` and `veg` packages get the same per-face, solid-boundary and instance-count reporting the other packages already have.

## What I've done

- New `04-frn-veg` workflow: surface checks, L14 solid boundary checks, per feature-type instance counts, and the per-file result sheet.
- `ShellManifold` now reports the edges that leave a shell open instead of the whole shell, which drives a new `非水密立体の辺` GeoJSON layer.
- Added that edge layer to `02-bldg` as well, and renamed `非水密体` to `非水密立体` across plateau6 to match the original workspace.

## What I haven't done

- `L-frn-01` / `L-frn-02` are not ported: they were dropped from the quality requirements, and the README now records that instead of the old porting notes.

## How I tested

New workflow tests under `testing/data/testcases/quality-check/plateau6/04-frn-veg/` (no-error, instance counts, surface error, solid error, vegetation, common `C01`), plus unit tests for the new edge positions.

## Memo

- `ShellConnected`, `Orientable` and `ShellOrientation` still report the whole shell, so for those the issue-location layer repeats `立体のエラー`. Kept in rather than dropped, so no finding is missing from the layer.
- `extractTags` stays empty: neither `frn` nor `veg` nests child city objects, so nothing is left behind.


shunski (approved):
👍

PR #2478, opened 2026-09-14
fix(engine): add cesium writer `properties` and fix PLATEAU6 attributes

asrcpq:
# Overview

This PR adds tileset properties for tileset-level attribute/range display. It also implements and fixes many attributes inconsistencies in the two implemented workflows in PLATEAU6.

## Changes

- implement `properties` in tileset.json by collecting attributes when processing output features, add 01-bldg/fld test for it.
- implement disaster (fld, lsld, etc.) attributes extraction
- remove attribute key sanitization in new-geometry cesium tiles metadata writer. The same problem as #1874.
- implement city_code and meshcode attributes to match PLATEAU4
- add error message when table extractor coercion fails.
- swap _x/_y attribute order to match PLATEAU convention
- fix several flattening rules

## Notes

- it is still an open question if we should extend cesium metadata to support unicode.
- mvt and cesium sanitization is currently inconsistent (same as PLATEAU4), need further discussion.


n4to4 (approved):
👍

PR #2481, opened 2026-09-15
feat(engine): add the plateau6 05-luse-urf-area quality check

n4to4:
# Overview

Ports the 05-luse-urf-area quality check to PLATEAU6 (CityGML 3.0 + i-UR 4.0), covering the `luse`, `urf` and `area` packages.

## What I've done

- Run the per-face geometry checks, the face orientation check and the per feature-type instance count over all three packages
- Emit per-package result files, error GeoJSON layers and `summary_<package>.json`, with the prefix picked from a `packagePrefixes` table
- Register the `area` package in `domain_of_definition_validator`, which had no entry for it

## What I haven't done

- Only the new geometry world is supported; this workflow is not expected to run under the legacy geometry types.

## How I tested

Six `workflow_test.json` cases under `testing/data/testcases/quality-check/plateau6/05-luse-urf-area/`: a no-error case, instance counts, orientation, and one surface-error case per package.

- The `packagePrefixes` table aborts on an unlisted package rather than letting two packages share an output file — is that the behaviour you want?

## Memo

- Output naming, column order and the 検査結果一覧 row shape follow the original workspace, not the plateau4 port.
- The checks live in a separate `luse_urf_area_checks.yml` sub-graph instead of one large `workflow.yml` as in plateau4, matching the recent PLATEAU6 workflows.
- Reprojection runs before every check because planarity and absolute winding are only defined in a linear-unit frame; a city object that fails to reproject is reported as a surface error rather than silently dropped.


shunski (approved):
👍

PR #2484, opened 2026-09-15
feat: jgd2024 support for plateau workflows

shunski:
# Overview
This PR implements vertical reprojections around jgd2024, and modifies the PLATEAU 4 DC workflows so that they can create a view that shows features at revised JGD2024 height, even if the input data is created in JGD2011. This PR builds on top of [the nusamai crate change](https://github.com/reearth/plateau-gis-converter/pull/81).

## Changes to Vertical Reprojector
- It now supports a new mode `Jgd2024ToWgs84`, which will be used to convert Jgd2024 height to Wgs84. When the data is given with the Jgd2011 (epsg 6697) height, it revise the height to that of Jgd2024 first, and then convert it to Wgs84.
- User can configure what the converter do when it encounters the coverage hole of the revision data. It either passes though or raises an error. On the "pass-through" option, number of conversions that are skipped are recorded and emitted at the end as a warning.

## Workflow changes
- Every workflow changes the VerticalReprojector's parameter to `Jgd2024ToWgs84`. It uses the new mode `Jgd2024ToWgs84`, with the default "pass-through" option for the coverage hole.
- The 09-unf workflow's HorizontalReprojector changes the target epsg from 4326 to 6697, since the downstream VerticalReprojector now accepts 6697 and 11318 only. There is no behavioral change on the HorizontalReprojector, because the horizontal part of reprojection is the same for 4326 and 6697.

## Note
- First, I tried to implement the epsg 6697 / 11318 separation inside the workflow, not inside the Vertical Reprojector, so that the behavior is explicit from what's in the workflow. However, the epsg code was not stored inside attributes, but instead it's recorded inside geometry, so the `FeatureFilter` routing couldn't be used for this purpose.


n4to4 (approved):
👍

PR #2486, opened 2026-09-16
fix(engine): implement cesium metadata type and fix feature welding

asrcpq:
# Overview

This PR implements value types in cesium writer. Previously, all cesium attribute values were converted to strings in tileset such as [this truth file](https://github.com/reearth/reearth-flow/pull/2486/changes#diff-6dcb60a18abba6f32874eef1c469769bd7d078f9a2ed7dd0aa1a91356a63d262).

This PR also fixes a pre-existing bug that feature gets vertices welded, by using feature id as the dedup key. [A unit test](https://github.com/reearth/reearth-flow/pull/2486/changes#diff-7e7cd924fdf615a94f48642d32adfd41bffab753181fcec12e02b69f4501483aR302-R343) is added.

## Changes

- implement metadata column type inference by using the most general type (int->float->string) to hold all values in a glb. This means attribute value type of some key is expected to be consistent across tileset if type matters.
- add feature ID to dedup key and remove currently empty vertex attribute.
- update all tests to fix attribute types in glb
- fix texture loading for extension textures (needed by PLATEAU5 branch)
- fix padding. Structural metadata buffer views uses correct bytewidth for padding. JSON BIN chunk aligns to 8.

## References

- Buffer view padding requirement: https://github.com/CesiumGS/glTF/blob/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata/README.md?plain=1#L277
- 8-byte JSON chunk extra padding: https://github.com/CesiumGS/glTF/blob/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata/README.md?plain=1#L523


n4to4 (approved):
LGTM 👍

PR #2490, opened 2026-09-17
feat(engine): add draco quantization param

shunski:
# Overview
This PR adds quantization parameter for the Cesium 3DTiles Writer and Gltf Writer, so that user can control the quantization errors.
Then it modifies the PLATEAU DC workflows so that it uses 0.01 m (=1 cm) for LOD less than or equal to 2, and 0.003 m (3 mm) for LOD 3. These error bounds are determined by some experiments on actual PLATEAU datasets, but they are still experimental and may be changed in the future.

## Changes
- Cesium 3DTiles Writer and GltfWriter now have a richer config; previously it was a simple boolean for toggling enable/disable, but now the enable mode carries an optional parameter that specifies the quantization error bound.
- Though not exposed as a parameter, texture coordinate quantization bits have also changed. The actual bits are determined by the texture images' pixel size, and the error is bounded to half a texel.
- Tests surfaced the string offset bug in the Cesium writer, which was causing the gltf unloadable for some strict viewers such as CesiumJS. It is fixed in this PR, though it is not directly related.

## Screenshots
Tsukuba brid LOD3 (left = before fix, right = after fix) 
Screenshot from 2026-09-16 15-50-17
Tsukuba tran LOD3 (before fix)
Screenshot from 2026-09-16 13-49-50
Tsukuba tran LOD3 (after fix)
Screenshot from 2026-09-16 13-49-57


n4to4 (approved):
👍

PR #2496, opened 2026-09-18
chore(engine): add gltf generator field

shunski:
This PR makes gltf writer fill the generator field. Previously the generator field was not specified.


n4to4 (approved):
👍

